Issue #62, opened 2014-10-09
Adding support for Privileged Mode


minid33:
It'd be pretty useful to remove the dependency on the external containerizer for docker in docker images which require privileged mode. 

To do this we mesos needs to complete this pull request:
https://github.com/apache/mesos/pull/26

I'm looking into adding this to jenkins mesos-plugin.

PR #69, opened 2014-11-10
Privileged mode

minid33:
https://github.com/jenkinsci/mesos-plugin/issues/62

As promised here is the support for privileged mode. This was recently added by @tnachen to mesos: https://reviews.apache.org/r/26896/


src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -180,13 +180,15 @@ public String getImage() {
     private final String type;
     private final String dockerImage;
     private final List<Volume> volumes;
+    private final Boolean dockerPrivilegedMode;
```
KostyaSha: Why not primitive?

vinodkone:
Can you address the outstanding comments and rebase? Would love to get this into next release.

src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
```diff
@@ -92,6 +92,9 @@
                                     <f:entry title="${%Docker Image}" field="dockerImage" description="If using Docker, specify the docker image.">
                                         <f:textbox field="dockerImage" clazz="required" value="${slaveInfo.containerInfo.dockerImage}" default=""/>
                                     </f:entry>
+                                    <f:entry title="${%Docker Privileged Mode}" field="dockerPrivilegedMode" description="This will start the image using Docker's privileged mode.">
+                                        <f:checkbox field="dockerPrivilegedMode" value="${slaveInfo.containerInfo.dockerPrivilegedMode}" checked="false" />
```
KostyaSha: You don't need define field twice

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -180,13 +180,15 @@ public String getImage() {
...
 
     @DataBoundConstructor
-    public ContainerInfo(String type, String dockerImage, List<Volume> volumes)
+    public ContainerInfo(String type, String dockerImage, List<Volume> volumes, Boolean dockerPrivilegedMode)
       throws FormException {
```
KostyaSha: Usually IDE indent throw on new line left than internal code block to be more readable.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -410,7 +410,8 @@ public boolean configure(StaplerRequest request, JSONObject object)
 
               containerInfo = new MesosSlaveInfo.ContainerInfo(
                   containerInfoJson.getString("type"),
-                  containerInfoJson.getString("dockerImage"), volumes);
+                  containerInfoJson.getString("dockerImage"), volumes,
+                  containerInfoJson.getBoolean("dockerPrivilegedMode"));
```
KostyaSha: Don't understand this style. Either width not enough, either volumes should be on new line...

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
barkerd427: Looks like other blocks add an extra two spaces than what is here.

src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
barkerd427: Also, does this setting maintain the checked value appropriately? I remember another PR doing this differently to maintain the checked value. You can test this by saving the checked state and then opening the config and seeing if the field is still checked. There was an issue with the volumes field for read only.

KostyaSha: Open config, set checkbox, save, check xml config on fs, then open this page again and review checkbox (in case of wrong fatabinding). Also, why checked is need to be defined at all? There is field, there is default option and there is checkbox default itself.

barkerd427: https://github.com/jenkinsci/mesos-plugin/pull/74/files

maselvaraj:
@minid33 Thanks for the PR! 
https://github.com/jenkinsci/mesos-plugin/pull/123 is merged which contains docker privileged mode implementation based on your PR. So I am closing this.

PR #70, opened 2014-11-13
Fix for race condition during scheduler registration when there are multiple builds on queue

maselvaraj:
Fix for race condition during scheduler registration(on demand) when there are multiple builds on queue


cornelf:
@vinodkone good stuff there. We are running 0.5.0 version of the mesos-plugin and we are having the same issue. Any chance for a 0.5.1 release soon to include this change as well?

tomstockton:
:+1: +1  0.5.1. release with this fix would be excellent!!

maselvaraj:
@cornelf @tomstockton 0.6.0 was released today which includes this fix.

PR #73, opened 2014-12-28
[JENKINS-26236] Add Docker parameter functionality

barkerd427:
This will allow the use of Docker parameters which is available in mesos 0.21.0. This also brings the plugin up-to-date with the latest mesos release. My use-case has been to allow the sharing of data containers among slaves on a particular node. Specifically not requiring tools and maven dependencies to have long download times for each build. There are likely other use-cases that I have yet to explore, but this has been tested in my Jenkins instance.


vinodkone:
This is awesome!! Thanks for the PR.

PR #74, opened 2015-01-07
[JENKINS-26321] Booleans for isReadOnly (Volume), isExtract and isExecutable (URI) are not handled properly

ederst:
Please refer to https://issues.jenkins-ci.org/browse/JENKINS-26321 for details.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
@@ -107,8 +107,8 @@
                                                 <f:textbox field="hostPath" default="" value="${volume.hostPath}" />
                                             </f:entry>
 
-                                            <f:entry title="${%Read-Only}">
-                                                <f:checkbox field="readOnly" default="false" value="${volume.readOnly}" />
+                                            <f:entry title="${%Read-Only}" field="volume.readOnly">
+                                                <f:checkbox default="false" checked="${volume.readOnly}" />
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
```diff
@@ -107,8 +107,8 @@
                                                 <f:textbox field="hostPath" default="" value="${volume.hostPath}" />
                                             </f:entry>
 
-                                            <f:entry title="${%Read-Only}">
-                                                <f:checkbox field="readOnly" default="false" value="${volume.readOnly}" />
+                                            <f:entry title="${%Read-Only}" field="volume.readOnly">
+                                                <f:checkbox default="false" checked="${volume.readOnly}" />
```
KostyaSha: Why do you need "checked" value?

ederst: The "checked" attribute is for the checkbox to show whether the saved value is true or false. Attribute "value" would not show the actual saved value for me.

This means when, lets say, "readOnly" was saved as "true" the GUI would show it as unchecked and therefore "false".

Consequently this also means, when someone saves it again, it would override the "true" value in the config with the "false" value (without actively changing the value, but changing something else in the config, like a Maven Installation or something...).

I hope this is somewhat understandable (if not I can also provide screenshots).

barkerd427:
This has been annoying me. Big +1!

PR #75, opened 2015-01-07
[JENKINS-26324] Add configurable bridged network mode for Docker containers

ederst:
This pull request adds a new feature described in https://issues.jenkins-ci.org/browse/JENKINS-26324

Changes made:
* ~~Updated Mesos libs to 0.21.0 and HPI plugin to 1.106~~
* HOST and BRIDGE networking for docker containers can now be configured
* Port mappings for those are also possible (if Host Port is left empty, a random Port on the host is used)

~~Snuck in changes (I tried to keep commits atomic, but this happened anyway):~~
* ~~Slave name on Jenkins is now dependent on label (which might not always be good); See: String name = slaveInfo.getLabelString() + "-" + UUID.randomUUID().toString();~~
* ~~The hostname of the docker container is set to the name of the slave~~
* ~~An Exception is now logged if something happens in createMesosTask()~~
* ~~HUDSON_HOME changed from jenkins to /home/jenkins at SLAVE_COMMAND_FORMAT~~

Also: The code for the evaluation of the port range received by the offer is a bit dodgy. I'll try to rewrite this if I got some spare time.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -119,6 +119,7 @@
               <artifactId>maven-compiler-plugin</artifactId>
               <configuration>
                   <target>1.6</target>
+                  <source>1.6</source>
               </configuration>
           </plugin>
       </plugins>
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -28,12 +28,11 @@ import java.util.Map;
 import java.util.Queue;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
-
-import com.google.protobuf.ByteString;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.mesos.MesosSchedulerDriver;
@@ -41,6 +40,8 @@ import org.apache.mesos.Protos.Attribute;
 import org.apache.mesos.Protos.CommandInfo;
 import org.apache.mesos.Protos.ContainerInfo;
 import org.apache.mesos.Protos.ContainerInfo.DockerInfo;
+import org.apache.mesos.Protos.ContainerInfo.DockerInfo.Network;
+import org.apache.mesos.Protos.ContainerInfo.DockerInfo.PortMapping;
 import org.apache.mesos.Protos.Credential;
 import org.apache.mesos.Protos.ExecutorID;
 import org.apache.mesos.Protos.Filters;
@@ -57,12 +58,13 @@ import org.apache.mesos.Protos.TaskID;
 import org.apache.mesos.Protos.TaskInfo;
 import org.apache.mesos.Protos.TaskStatus;
 import org.apache.mesos.Protos.Value;
-import org.apache.mesos.Protos.Value.Type;
+import org.apache.mesos.Protos.Value.Range;
 import org.apache.mesos.Protos.Volume;
 import org.apache.mesos.Protos.Volume.Mode;
 import org.apache.mesos.Scheduler;
 import org.apache.mesos.SchedulerDriver;
 
+import com.google.protobuf.ByteString;
 
 public class JenkinsScheduler implements Scheduler {
   private static final String SLAVE_JAR_URI_SUFFIX = "jnlpJars/slave.jar";
@@ -147,7 +149,7 @@ public class JenkinsScheduler implements Scheduler {
     if (driver != null) {
       driver.stop();
     } else {
-	  LOGGER.warning("Unable to stop Mesos driver:  driver is null.");
+      LOGGER.warning("Unable to stop Mesos driver:  driver is null.");
     }
     running = false;
   }
@@ -247,7 +249,12 @@ public class JenkinsScheduler implements Scheduler {
         if (matches(offer, request)) {
           matched = true;
           LOGGER.info("Offer matched! Creating mesos task");
-          createMesosTask(offer, request);
+
+          try {
+              createMesosTask(offer, request);
+          } catch (Exception e) {
+              LOGGER.log(Level.SEVERE, e.getMessage(), e);
+          }
           requests.remove(request);
           break;
         }
@@ -262,6 +269,7 @@ public class JenkinsScheduler implements Scheduler {
   private boolean matches(Offer offer, Request request) {
     double cpus = -1;
     double mem = -1;
+    List<Range> ports = null;
 
     for (Resource resource : offer.getResourcesList()) {
       if (resource.getName().equals("cpus")) {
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -62,18 +62,15 @@
   </pluginRepositories>
 
   <properties>
-    <!--
-      explicitly specifying the latest version here because one we get from the parent POM
-      tends to lag behind a bit
-    -->
-    <maven-hpi-plugin.version>1.95</maven-hpi-plugin.version>
+    <maven-hpi-plugin.version>1.106</maven-hpi-plugin.version>
```
KostyaSha: imho you can update jenkins core version to 1.565.3 that is previous LTS branch version and contains exactly the same hpi-plugin version.

KostyaSha:
Slow logging everywhere.

pom.xml
ederst: Hmm, yeah. This change was not really meant to be in there as well (sloppy job of me reviewing the changes). I think it is not really necessary for this feature to work. However, I have to test this and clean up my pull request a bit.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -284,6 +308,7 @@ private boolean matches(Offer offer, Request request) {
           "\nRequested for Jenkins slave:\n" +
           "  cpus: " + requestedCpus + "\n" +
           "  mem:  " + requestedMem + "\n" +
+          "  ports: " + hasPortMappings + "\n" +
```
barkerd427: Might be more helpful to know what ports are being requested.

src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
```diff
@@ -95,6 +95,37 @@
                                 </f:radioBlock>
                             </f:entry>
 
+                            <f:entry title="${%Networking}" description="${%Specify the networking mode to use for this container}" field="networking" >
+                                <f:radioBlock id="networking.HOST" name="networking" title="${%Host}" value="HOST" inline="true" checked="${slaveInfo.containerInfo.networking == 'HOST'}" />
```
barkerd427: What about the `container` option?

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -180,13 +184,27 @@ public String getImage() {
...
       this.dockerImage = dockerImage;
       this.volumes = volumes;
+
+      if (networking == null) {
+          this.networking = Network.BRIDGE.toString();
+      } else {
+          this.networking = networking;
+      }
+
+      if (Network.HOST.equals(Network.valueOf(networking))) {
```
barkerd427: Should a PortMapping perhaps set the host port if the networking is HOST as the mapping is really just implied in Docker?

src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
barkerd427: Disregard. I hadn't realized it wasn't included in Mesos.

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -10,8 +11,9 @@
 import net.sf.json.JSONSerializer;
 
 import org.apache.commons.lang.StringUtils;
-import org.apache.mesos.Protos.ContainerInfo;
+import org.apache.mesos.Protos.ContainerInfo.DockerInfo.Network;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.springframework.transaction.annotation.Transactional;
```
barkerd427: Why was this added?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -391,11 +444,63 @@ private void createMesosTask(Offer offer, Request request) {
...
+              List<Integer> portsToUse = findPortsToUse(offer, request, portMappings.size());
+
+              Value.Ranges.Builder portRangesBuilder = Value.Ranges.newBuilder();
+
+              for (MesosSlaveInfo.PortMapping portMapping : portMappings) {
+                  PortMapping.Builder portMappingBuilder = PortMapping.newBuilder() //
+                          .setContainerPort(portMapping.getContainerPort()) //
+                          .setProtocol(portMapping.getProtocol());
+
+                  int portToUse = portMapping.getHostPort() == null?portsToUse.get(portToUseIndex++):portMapping.getHostPort();
```
barkerd427: Use spaces around the operators.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -71,7 +75,7 @@
   private static final double JVM_MEM_OVERHEAD_FACTOR = 0.1;
 
   private static final String SLAVE_COMMAND_FORMAT =
-      "java -DHUDSON_HOME=jenkins -server -Xmx%dm %s -jar ${MESOS_SANDBOX-.}/slave.jar %s -jnlpUrl %s";
+      "java -DHUDSON_HOME=/home/jenkins -server -Xmx%dm %s -jar ${MESOS_SANDBOX-.}/slave.jar %s -jnlpUrl %s";
```
KostyaSha: Why /home/?

vinodkone:
Have all the comments raised so far been addressed? Also, can you please rebase?

ederst:
@vinodkone Hi, will address the comments (and clean the pull request) when I have some spare time. Right now I'm quite busy.

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
ederst: Oversight, does not need to be there...

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
ederst: @barkerd427 I'm not sure if I understand you correctly, but if I do then you suggest, adding a port mapping configuration for the HOST network mode, just for the host port?! I don't know if this is allowed by Docker (will have to try this).

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
ederst: @barkerd427 True that. Realized this when I started a slave with the VNC port exposed and I had no clue to which host port I have to connect. In the meantime I've implemented a feature which shows the (host) ports exposed on the Mesos Slaves on the Jenkins Slave Status page. Still, I have to test this and prepare another pull request...

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
ederst: @barkerd427 AFAIK now is that HOST mode in Docker exposes the ports of all services in a container on the host network interface. This means that it will not be possible to map ports differently (the ports would have to be configured in the service configuration, for example "Listen 80" in httpd.conf)

barkerd427: You're right. I was confused it seems.

ederst:
I'd say all issues are addressed so far, so the pull request can be merged. If not, please tell me why and I'll try to fix it.

maselvaraj:
@ederst Can you please squash the commits ?

ederst:
@maselvaraj Squashed

Issue #76, opened 2015-01-08
Maven build in Vagrant instance fails


hardieas:
Vagrant up and vagrant ssh work but running mvn hpi:run after ssh-ing to vagrant instance fails with this error:

[INFO] --- maven-hpi-plugin:1.95:run (default-cli) @ mesos ---
[INFO] Generating ./work/plugins/mesos.hpl
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 11.487s
[INFO] Finished at: Thu Jan 08 14:19:52 UTC 2015
[INFO] Final Memory: 20M/230M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.jenkins-ci.tools:maven-hpi-plugin:1.95:run (default-cli) on project mesos: Error preparing the manifest: ./work/plugins/mesos.hpl (No such file or directory) -> [Help 1]

KostyaSha:
Please provide exact command and its result.

PR #80, opened 2015-01-15
Upgraded to build with Mesos 0.21.1

nickithewatt:
Upgraded to build with latest versions of Mesos 0.21.1 and Jenkins LTS (at least the Long Term Support release) 1.580.2.
Required some updating to Vagrantfile as well, which additionally resulted in fix for issue #76, plus now has the option to use the local plugin source instead of explicit release version if desired within the VM. This is helpful for some testing in VM, but as per previous code still defaults to released version of plugin (now latest version as well - 0.5.0)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.509.1</version><!-- which version of Jenkins is this plugin built against? -->
+    <version>1.565.3</version><!-- which version of Jenkins is this plugin built against? -->
   </parent>
 
   <groupId>org.jenkins-ci.plugins</groupId>
@@ -41,7 +41,6 @@
       <email>vinodkone@gmail.com</email>
     </developer>
   </developers>
-  
   <contributors>
     <contributor>
       <name>Daniel Barker</name>
@@ -74,18 +73,20 @@
       tends to lag behind a bit
     -->
     <maven-hpi-plugin.version>1.95</maven-hpi-plugin.version>
+    <mesos.version>0.21.1</mesos.version>
+    <protobuf.version>2.5.0</protobuf.version>
   </properties>
 
   <dependencies>
       <dependency>
           <groupId>org.apache.mesos</groupId>
           <artifactId>mesos</artifactId>
-          <version>0.21.0</version>
+          <version>${mesos.version}</version>
       </dependency>
       <dependency>
           <groupId>com.google.protobuf</groupId>
           <artifactId>protobuf-java</artifactId>
-          <version>2.5.0</version>
+          <version>${protobuf.version}</version>
       </dependency>
   </dependencies>
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,14 @@
 target/
 work/
 
+# Exclude IntelliJ files
+.idea/
+*.iml
+*.ipr
+
+# Exclude Vagrant generated files
+.vagrant/
+
 # Used as a local build/deploy dev-cycle script
 deploy.sh
 
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ This should build the Mesos native library in the `build/src/.libs` folder.
 
 ### Vagrant ###
 
-If you are just looking to play with Mesos and this plugin in a VM, you could do so with the included Vagrantfile.
+If you are just looking to play with Mesos and this plugin in a single self contained VM, you could do so with the included Vagrantfile.
 
 		$ vagrant up
 		$ vagrant ssh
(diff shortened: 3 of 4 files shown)


pom.xml
```diff
@@ -3,7 +3,9 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.509.1</version><!-- which version of Jenkins is this plugin built against? -->
+    <version>1.580.2</version><!-- which version of Jenkins is this plugin built against? (Current LTS as at 14 Jan 2015)-->
```
KostyaSha: This value defines the minimal jenkins core on what this plugin can be run. Don't update it without any critical reason.
Such update cuts plugin usage for thousands installations.

pom.xml
```diff
@@ -66,19 +68,21 @@
...
+    <mesos.version>0.21.1</mesos.version>
+    <protobuf.version>2.5.0</protobuf.version>
   </properties>
 
   <dependencies>
       <dependency>
           <groupId>org.apache.mesos</groupId>
           <artifactId>mesos</artifactId>
-          <version>0.20.0</version>
+          <version>${mesos.version}</version>
```
KostyaSha: I see no reason to move this single values to variables.

pom.xml
nickithewatt: Fair enough, I had updated it because it did resolve an issue when building with Java 8, i.e. you got the following error:

    Tests in error:
          org.jvnet.hudson.test.JellyTestSuiteBuilder$JellyTestSuite(org.jvnet.hudson.test.junit.FailedTest): -1
    testCliSanity(org.jvnet.hudson.test.PluginAutomaticTestBuilder$CliSanityTest): second instance

however as you correctly point out, this is not critical and will work with JDK 7 so should not be changed, I shall revert this

pom.xml
nickithewatt: It does allow you to trial building the plugin with a new version of mesos without physically changing the pom, for example when a new version does came out and you want to trial building you could do the following:
    mvn clean package -Dmesos.version=0.22

With regards to generally changing these values to variables, that has simply been a standard I have tended to follow in pom files, I find it provides a nice concise way to see upfront (in the properties section) what all the main dependencies and versions are being used - however if you feel this is not beneficial, then I am happy to remove it as well, just let me know.

KostyaSha: It up to maintainers, but such builds will be not reproducible, hope plugin wasn't released in such way. If you want newer version then explicitly define. If you want update all versions to latest, then there is maven versions plugin.

pom.xml
```diff
@@ -66,19 +66,21 @@
       explicitly specifying the latest version here because one we get from the parent POM
       tends to lag behind a bit
     -->
-    <maven-hpi-plugin.version>1.95</maven-hpi-plugin.version>
+    <maven-hpi-plugin.version>1.106</maven-hpi-plugin.version>
```
KostyaSha: https://github.com/jenkinsci/mesos-plugin/pull/75#discussion_r23129793

pom.xml
nickithewatt: As per previous note, if this was used, it would only really be for local trialling of different mesos versions for experimentation. I would never expect it to be built and released this way

pom.xml
nickithewatt: @KostyaSha just seen your #75 comment. Does that not contradict your previous point about changing the minimal jenkins core version, or is there else something special going on in this case?

KostyaSha: Usually we try to keep 2y or 1y core, assuming that many plugins were updated for workflow plugin core requirements. Keeping 2 or 3 last lts versions should ok. I don't know any bugs that blocks people not to use 1.565.3, so imho 1.565.3 is ok.

vinodkone:
LGTM. Just one question about hpi-plugin version. What was the reason for the version bump?

barkerd427:
+1

KostyaSha:
@vinodkone there can be java8, databoundsetters or other things. My suggestion will be remove hpi update and bump core to 1.565.3

barkerd427:
I agree with @KostyaSha ^^

nickithewatt:
@vinodkone Yes as per @KostyaSha comments, I had originally bumped the jenkins version up to cater for being able to build in a Java 8 setup without errors (which I had) and I tried to align the hpi version with that as well as the same time. However there was a bit of backwards and forwards regarding whether this would limit existing setups or not. It was thus a convenience for those building in a Java 8 setup, however all still works with Java 7. As per suggestions happy to bump jenkins version and downgrade hpi.

nickithewatt:
Rebased, squash commits and adjusted as per comment, i.e. 
Mesos - 0.21.1
Jenkins - 1.565.3
HPI - 1.95

vinodkone:
Thanks for the cleanup. Merging it now.

PR #84, opened 2015-02-07
jnlp secret for authentication if security is enabled in jenkins

maselvaraj:


vinodkone:
Awesome. Thanks!

bacoboy:
This is great.  Thanks for taking point...

Issue #71, opened 2014-12-10
Failed to redirect stdout


jc8390:
Hi,
I am running mesos 0.20.1, jenkins 1.593 and jenkins-mesos 0.5.0

I can see the framework is registered successfully with cluster and connection between jenkins server and mesos cluster is good. 
When I am trying to execute jenkins jobs but tasks are failed. 
This is the error I am seeing in the mesos slave logs.

"Container 'bbc2d6d4-aeae-4b16-ae4e-28801d78a8b0' for executor 'mesos-jenkins-2e83dadd-f20d-4ad4-b051-11b1abe56623' of framework '20141210-203527-16777343-5050-2717-0000' failed to start: Failed to redirect stdout: Failed to chown: Failed to execute 'chown -R 115:125 '/tmp/mesos/slaves/20141210-203527-16777343-5050-2717-1/frameworks/20141210-203527-16777343-5050-2717-0000/executors/mesos-jenkins-2e83dadd-f20d-4ad4-b051-11b1abe56623/runs/bbc2d6d4-aeae-4b16-ae4e-28801d78a8b0/stdout'' (exit status: 256): Success"

Launching task mesos-jenkins-2fabf12e-d79f-440b-a5db-5f82777dde36 for framework 20141210-203527-16777343-5050-2717-0000
I1210 20:53:19.602452  3495 slave.cpp:3857] Checkpointing ExecutorInfo to '/tmp/mesos/meta/slaves/20141210-203527-16777343-5050-2717-1/frameworks/20141210-203527-16777343-5050-2717-0000/executors/mesos-jenkins-2fabf12e-d79f-440b-a5db-5f82777dde36/executor.info'
I1210 20:53:19.603184  3493 containerizer.cpp:394] Starting container '40eb9964-0ef3-428a-98c0-82d2326a47fc' for executor 'mesos-jenkins-2fabf12e-d79f-440b-a5db-5f82777dde36' of framework '20141210-203527-16777343-5050-2717-0000'
[...]

PR #92, opened 2015-03-11
Added an option to specify Jenkins URL. Also a minor bug fix for on dema...

maselvaraj:
...nd registration

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
@@ -62,6 +62,7 @@ public class MesosCloud extends Cloud {
   private String secret;
   private final boolean checkpoint; // Set true to enable checkpointing. False by default.
   private boolean onDemandRegistration; // If set true, this framework disconnects when there are no builds in the queue and re-registers when there are.
+  private String jenkinsURL;
 
   // Find the default values for these variables in
   // src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly.
@@ -121,7 +122,8 @@ public class MesosCloud extends Cloud {
       String secret,
       List<MesosSlaveInfo> slaveInfos,
       boolean checkpoint,
-      boolean onDemandRegistration) throws NumberFormatException {
+      boolean onDemandRegistration,
+      String jenkinsURL) throws NumberFormatException {
     super("MesosCloud");
 
     this.nativeLibraryPath = nativeLibraryPath;
@@ -134,12 +136,14 @@ public class MesosCloud extends Cloud {
     this.slaveInfos = slaveInfos;
     this.checkpoint = checkpoint;
     this.onDemandRegistration = onDemandRegistration;
-
-    JenkinsScheduler.SUPERVISOR_LOCK.lock();
-    try {
-      restartMesos();
-    } finally {
-      JenkinsScheduler.SUPERVISOR_LOCK.unlock();
+    this.setJenkinsURL(jenkinsURL);
+    if(!onDemandRegistration) {
+	    JenkinsScheduler.SUPERVISOR_LOCK.lock();
+	    try {
+	      restartMesos();
+	    } finally {
+	      JenkinsScheduler.SUPERVISOR_LOCK.unlock();
+	    }
     }
   }
 
@@ -158,6 +162,14 @@ public class MesosCloud extends Cloud {
       nativeLibraryLoaded = true;
     }
 
+    // Default to root URL in Jenkins global configuration.
+    String jenkinsRootURL = Jenkins.getInstance().getRootUrl();
+
+    // If 'jenkinsURL' parameter is provided in mesos plugin configuration, then that should take precedence.
+    if(StringUtils.isNotBlank(jenkinsURL)) {
+      jenkinsRootURL = jenkinsURL;
+    }
+
     // Restart the scheduler if the master has changed or a scheduler is not up.
     if (!master.equals(staticMaster) || !Mesos.getInstance().isSchedulerRunning()) {
       if (!master.equals(staticMaster)) {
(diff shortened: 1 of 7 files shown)


nemothekid:
Built this branch, works thus far

maselvaraj:
@nemothekid Thanks for testing this feature!

daniel-beck:
Jenkins advertises the header `X-Jenkins-CLI-Host` on its home page when the system property `hudson.TcpSlaveAgentListener.hostName` is set. I think for a similar reason. Can't you just reuse that?

maselvaraj:
@daniel-beck  Thanks for the comment. But I believe hudson.TcpSlaveAgentListener.hostName works only for CLI. I can see a bug logged here : https://issues.jenkins-ci.org/browse/JENKINS-27218 which claims that the header is not used by slave.jar.

daniel-beck:
If set, it specifies the actual host Jenkins runs on. So it could easily be reused by this plugin instead of adding another option.

maselvaraj:
@daniel-beck How about the port number ?

daniel-beck:
True. It will not tell you the port Jenkins runs on, as the CLI connects to a different port. Good point.

nemothekid:
Hmm, this was working when I posted but now it doesn't. New tasks cannot be started and they all fail with:

```
WARNING: Logging before InitGoogleLogging() is written to STDERR
I0318 17:45:28.406891 29052 fetcher.cpp:76] Fetching URI 'jenkins.marathon.mesos:8888/jnlpJars/slave.jar'
E0318 17:45:28.407048 29052 fetcher.cpp:161] A relative path was passed for the resource but the environment variable MESOS_FRAMEWORKS_HOME is not set. Please either specify this config option or avoid using a relative path
Failed to fetch: jenkins.marathon.mesos:8888/jnlpJars/slave.jar
Failed to synchronize with slave (it's probably exited)
```

nemothekid:
After looking at the mesos source code, it looked like mesos thought the url was a local path. I had to change the jenkins url from `jenkins.marathon.mesos:8888` to `http://jenkins.marathon.mesos:8888`. The plugin should either fill in `http://` or the help message should make this clear (it currently says to just use `host:port`

maselvaraj:
@nemothekid Thanks. That's a good point. I will update the help file.

src/main/java/org/jenkinsci/plugins/mesos/MesosComputer.java
```diff
@@ -43,4 +50,10 @@ public HttpResponse doDoDelete() throws IOException {
       getNode().terminate();
     return new HttpRedirect("..");
   }
+
+  @Override
+  @WebMethod(name="slave-agent.jnlp")
+  public HttpResponse doSlaveAgentJnlp(StaplerRequest req, StaplerResponse res) throws IOException, ServletException {
```
vinodkone: What is this for?

src/main/resources/org/jenkinsci/plugins/mesos/MesosComputer/mesos-slave-agent.jnlp.jelly
```diff
@@ -0,0 +1,75 @@
...
+  <!--
+    See http://www.dallaway.com/acad/webstart/ for obtaining the certificate.
+  -->
+
+    <!-- See http://java.sun.com/j2se/1.5.0/docs/guide/javaws/developersguide/syntax.html for the syntax -->
+    <jnlp spec="1.0+"
+        codebase="${rootURL}computer/${h.encode(it.node.nodeName)}/">
+
+      <information>
+        <title>Slave Agent for ${it.displayName}</title>
```
vinodkone: What is this for?

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -134,12 +136,14 @@ public MesosCloud(
...
     this.checkpoint = checkpoint;
     this.onDemandRegistration = onDemandRegistration;
-
-    JenkinsScheduler.SUPERVISOR_LOCK.lock();
-    try {
-      restartMesos();
-    } finally {
-      JenkinsScheduler.SUPERVISOR_LOCK.unlock();
+    this.setJenkinsURL(jenkinsURL);
+    if(!onDemandRegistration) {
```
vinodkone: What was the bug here?

maselvaraj: Bug was this: Even If  'ondemand' registration is selected, Mesos plugin would connect(restartMesos) to Mesos master when the user clicks save in Jenkins global configuration

So, I just included the check if the user has enabled ondemand registration before restarting Schedular.

src/main/resources/org/jenkinsci/plugins/mesos/MesosComputer/mesos-slave-agent.jnlp.jelly
maselvaraj: This is the default jelly that’ll be used which passes the URL via JNLP response : https://github.com/maselvaraj/jenkins/blob/master/core/src/main/resources/hudson/slaves/SlaveComputer/slave-agent.jnlp.jelly#L71 
 
Please check line #71 (This is the defualt URL configured in Jenkins) and #85 (This is the URL used to fetch JNLP response. In our case it is the url configured in Mesos plugin). I’ve just used the same jelly file, just removed the Jenkins root URL part, so that it talks only to URL from request.

src/main/java/org/jenkinsci/plugins/mesos/MesosComputer.java
maselvaraj: To override the default slave-agent.jnlp.jelly with the new file mesos-slave-agent.jnlp.jelly which would enable slave.jar to talk to only the Jenkins URL configured in Mesos plugin configuration.

bhenderson:
Hi, I commented on the mailing list thread, but I thought I'd post here too. I'm setting jenkinsUrl, which is allowing the slave.jar to get downloaded, but it seems like the jnlp host:port is wrong. The jelly file supports a tunnel option on the launcher, but there is no way to set that from the plugin.

thoughts?

Issue #94, opened 2015-03-25
Only 1 Mesos cluster supported at a time


bacoboy:
I have 2 different mesos clusters (long story -- let's just say 'production' and 'pre-production').  I have configured both in the jenkins master under different 'Cloud' entries and created a slave for each (with different tags).  When I launch my job, both tagged jobs get launched on whichever cluster whose scheduler got launched first.

I looked at the code, and there appears to be a lazy instantiation of a single Mesos object here: https://github.com/jenkinsci/mesos-plugin/blob/master/src/main/java/org/jenkinsci/plugins/mesos/Mesos.java#L89

What I was expecting was that a slave configuration tagged under the specific MesosCloud object would launch a different scheduler (and hence a different slave in that cluster).

Rather than assuming there is only 1 cluster in MesosComputerLauncher here: https://github.com/jenkinsci/mesos-plugin/blob/master/src/main/java/org/jenkinsci/plugins/mesos/MesosComputerLauncher.java#L58
can't we create a different Mesos object depending which.  It probably involves passing a reference to the  MesosCloud to the MesosSlave/MesosSlaveInfo????

Thoughts?

vinodkone:
As you correctly diagnosed, currently the plugin only supports one Mesos Cloud at a time. But if connecting to multiple Mesos clusters from the same Jenkins master is a use case that you are interested in, we can think about adding that feature (PRs are welcome too :)). Is running multiple Jenkins masters, one for each Mesos cluster, not an option for you?

bacoboy:
Already submitted pull request.  Please look it over...

vinodkone:
Ah. I missed that PR. I will take a look. Thanks.

PR #96, opened 2015-03-25
Fix support for multiple Mesos Clouds

bacoboy:
This is to address issue 94:
https://github.com/jenkinsci/mesos-plugin/issues/94

The Mesos object was written to be a singleton
even though Jenkins supports defining more than one.

In my case, I was trying to control 2 different clusters
(with 2 different slave tags).  Because the MesosImpl gets created by
whichever slave happens to launch first, slave tags from the other cluster
would launch in the wrong place.

While I don't LOVE this implementation, it seemed the least minimal
impact to a larger refactoring.  I basically pass a reference for the
MesosCloud to the MesosSlave.  Then I changed the Mesos.getInstance()
to take a parameter which is used as a key in a Map keyed by MesosCloud to
find the correct MesosImpl.  It was necessary to externalize the MesosImpl
inner-class to get it to compile.

I also cleaned up some unused imports, code not called, and depricated
use of Hudson for Jenkins.  You also can't override interfaces so I removed
some @Override tags

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ work/
 .idea/
 *.iml
 *.ipr
+*.iws
 
 # Exclude Vagrant generated files
 .vagrant/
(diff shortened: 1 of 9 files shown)


maselvaraj:
@bacoboy  Thanks. LGTM. Can you please squash the commits?

bacoboy:
Squashed.

maselvaraj:
Thanks for the PR!

bacoboy:
Cool.  Looking forward to the next official release :)

PR #98, opened 2015-04-16
Adding WebUI field to Framework Info

abhay-agarwal:
I added a web UI field so that Mesos registers the location of the Jenkins UI. If this is not available at runtime, then the WebUI will not register.


maselvaraj:
Thanks for the PR!

PR #101, opened 2015-04-24
Fix thread leak and deadlock on terminating slaves

ederst:
Thread Leak:
When no Task was created in Mesos (due to connection error to Mesos Master or no resources available) there was no "signal" sent to the waiting MesosSlaveLauncher to finish waiting. Therefore the thread count in the Java VM increased drastically.

Deadlock:
When terminating Slaves there is a possible deadlock when the Retention check starts. We observed and experienced this two times in the last weeks. Same issue (and solution) as here (https://issues.jenkins-ci.org/browse/JENKINS-22558), but different plugin.


maselvaraj:
Thanks for the PR!

PR #106, opened 2015-05-19
if we didn't have port mapping, don't try to grab them

DustinChaloupka:
Stopping the scheduler from blowing up when we didn't supply port mappings to be part of the resource offers check.


bacoboy:
This handles one of 2 NPEs.
See https://github.com/jenkinsci/mesos-plugin/issues/113 for my solution.
Basically I leave this line alone and guard deserializing null values in 2 places.  This makes sure the port mapping is an empty collection (so you won't see the NPE you are guarding for here).

I don't think we need both of these -- but it doesn't hurt to double-check.  Esp for debugging code -- which is what this is...

Now if we can just get this looked over quickly and a 0.7.1 released....

DustinChaloupka:
[https://github.com/jenkinsci/mesos-plugin/pull/117](That) works for me!

PR #114, opened 2015-06-02
reduce logging level for some status things

bacoboy:
Lowering the logging level from INFO to FINE for a few items.
The noisiest one is the 'not matching slave' message in mesos environments
where only a few of the slaves are jenkins candidates.  The non-matching ones
dump many large json objects into the logs which really falls more into
the debugging camp hence the lower level


bacoboy:
going to move to master branch

PR #119, opened 2015-06-06
Fix infinite loop in the port mapping code

stephenliang:
There's an infinite loop if the offer provides a single port mapping, where the begin = 31000 and the end = 31000. The inner while loop will never execute causing an infinite loop which effectively blocks JenkinsScheduler and causes a lock up in Jenkins

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ deploy.sh
 */target/*
 */build/*
 /bin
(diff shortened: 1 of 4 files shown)


maselvaraj:
@stephenliang Thanks for the PR. Awesome job with the unit test!

PR #121, opened 2015-06-11
Support one off slaves on the job level for mesos slaves.

### LINKED ISSUE #48: Have an option to only allow a slave to run one job and then terminate
Is it possible to have the Jenkins plugin only spin up a slave and use it for one run of a job and then kill the slave after the run is finished so that the slave only is used once? We're testing out the containerizer features and only want a container to live for one run of a job. I've seen this feature in other Jenkins plugins where the framework checks if a job has been run on a slave and if a job has run it kills the slave and spins up another one.

stephenliang:
Fixes #48

Based off the JClouds Jenkins plugin's implementation. The slave, on teardown, will mark temporarily offline, and thus no longer take any jobs. The Mesos retention strategy will then set it as pending deletion after the idle period to protect against post-build actions from terminating the slave too soon. The final step is the asynchronous period work in MesosCleanupTask which will terminate the pending deletion slave.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,6 +83,7 @@
     <mockito.version>1.10.19</mockito.version>
     <powermock.version>1.6.2</powermock.version>
     <assertj.version>2.1.0</assertj.version>
+    <guava.version>18.0</guava.version>
   </properties>
 
   <dependencies>
@@ -124,6 +125,11 @@
           <artifactId>assertj-core</artifactId>
           <version>${assertj.version}</version>
           <scope>test</scope>
+      </dependency>
+      <dependency>
+          <groupId>com.google.guava</groupId>
+          <artifactId>guava</artifactId>
+          <version>${guava.version}</version>
       </dependency>
   </dependencies>
 
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -17,6 +17,7 @@ package org.jenkinsci.plugins.mesos;
 
 
 import com.google.common.annotations.VisibleForTesting;
+import hudson.model.Computer;
 import hudson.model.Node;
 
 import java.util.ArrayList;
@@ -435,7 +436,28 @@ public class JenkinsScheduler implements Scheduler {
         refuseOffer(offer);
         return;
     }
-      
+
+    for (final Computer computer : Jenkins.getInstance().getComputers()) {
+        if (!MesosComputer.class.isInstance(computer)) {
+            LOGGER.finer("Not a mesos computer, skipping");
+            continue;
+        }
+
+        MesosComputer mesosComputer = (MesosComputer) computer;
+
+        if (mesosComputer == null) {
+            LOGGER.fine("The mesos computer is null, skipping");
+            continue;
+        }
+
+        MesosSlave mesosSlave = mesosComputer.getNode();
+
+        if (taskId.getValue().equals(computer.getName()) && mesosSlave.isPendingDelete()) {
+            LOGGER.info("This mesos task " + taskId.getValue() + " is pending deletion. Not launching another task");
+            driver.declineOffer(offer.getId());
+        }
+    }
+
     CommandInfo.Builder commandBuilder = getCommandInfoBuilder(request);
     TaskInfo.Builder taskBuilder = getTaskInfoBuilder(offer, request, taskId, commandBuilder);
 
@@ -454,6 +476,7 @@ public class JenkinsScheduler implements Scheduler {
   }
 
   private void detectAndAddAdditionalURIs(Request request, CommandInfo.Builder commandBuilder) {
+
     if (request.request.slaveInfo.getAdditionalURIs() != null) {
       for (MesosSlaveInfo.URI uri : request.request.slaveInfo.getAdditionalURIs()) {
         commandBuilder.addUris(
@@ -678,8 +701,7 @@ public class JenkinsScheduler implements Scheduler {
    */
   @VisibleForTesting
   void refuseOffer(Offer offer) {
-      Filters filters = Filters.newBuilder().setRefuseSeconds(1).build();
-      driver.launchTasks(offer.getId(), new ArrayList<TaskInfo>(), filters);
+      driver.declineOffer(offer.getId());
   }
 
   @Override
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosCleanupThread.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosCleanupThread.java
@@ -1,0 +1,115 @@
+package org.jenkinsci.plugins.mesos;
+
+import java.io.IOException;
+import java.util.logging.Level;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import com.google.common.util.concurrent.MoreExecutors;
+import hudson.Extension;
+import hudson.model.AsyncPeriodicWork;
+import hudson.model.Computer;
+import hudson.model.TaskListener;
+import jenkins.model.Jenkins;
+
+/**
+ This file is part of the JCloud Jenkins Plugin. (https://github.com/jenkinsci/jclouds-plugin)
+ commit id 20c6ca884abb172b27d0af82545c8915ce08f618.
+
+ This file was modified to work with the Mesos Jenkins plugin and based off the JClouds Jenkins Plugin.
+
+ According to the Jenkins Plugin guide (https://wiki.jenkins-ci.org/display/JENKINS/Before+starting+a+new+plugin),
+ If no license is defined in the form of a header, LICENSE file, or in the license section, the code is assumed to be
+ under The MIT License.
+
+ The MIT License (MIT)
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+
+ The above copyright notice and this permission notice shall be included in
+ all copies or substantial portions of the Software.
+
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ THE SOFTWARE.
+ */
+@Extension
+public class MesosCleanupThread extends AsyncPeriodicWork {
+
+    public MesosCleanupThread() {
+        super("Mesos pending deletion slave cleanup");
+    }
+
+    @Override
+    public long getRecurrencePeriod() {
+        return MIN * 1;
+    }
+
+    public static void invoke() {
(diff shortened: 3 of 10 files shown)


src/main/java/org/jenkinsci/plugins/mesos/MesosCleanupThread.java
```diff
@@ -0,0 +1,110 @@
...
+import com.google.common.util.concurrent.ListeningExecutorService;
+import com.google.common.util.concurrent.MoreExecutors;
+import hudson.Extension;
+import hudson.model.AsyncPeriodicWork;
+import hudson.model.Computer;
+import hudson.model.TaskListener;
+import jenkins.model.Jenkins;
+
+/**
+ This file is part of the JCloud Jenkins Plugin. (https://github.com/jenkinsci/jclouds-plugin)
```
barkerd427: Could we reference the commit from where this version initiated?

src/main/java/org/jenkinsci/plugins/mesos/MesosCleanupThread.java
```diff
@@ -0,0 +1,110 @@
...
+        final ImmutableList.Builder<ListenableFuture<?>> deletedNodesBuilder = ImmutableList.<ListenableFuture<?>>builder();
+        ListeningExecutorService executor = MoreExecutors.listeningDecorator(Computer.threadPoolForRemoting);
+        final ImmutableList.Builder<MesosComputer> computersToDeleteBuilder = ImmutableList.<MesosComputer>builder();
+
+        for (final Computer c : Jenkins.getInstance().getComputers()) {
+            if (MesosComputer.class.isInstance(c)) {
+                if (((MesosComputer) c).getNode().isPendingDelete()) {
+                    final MesosComputer comp = (MesosComputer) c;
+                    computersToDeleteBuilder.add(comp);
+                    logger.log(Level.INFO, "Marked " + comp.getName() + " for deletion");
```
barkerd427: nit: String.format("Marked %s for deletion.", comp.getName())

The same for other string formatting sections.

src/main/java/org/jenkinsci/plugins/mesos/MesosCleanupThread.java
```diff
@@ -0,0 +1,110 @@
...
+                    ListenableFuture<?> f = executor.submit(new Runnable() {
+                        public void run() {
+                            logger.log(Level.INFO, "Deleting pending node " + comp.getName());
+                            try {
+                                comp.getNode().terminate();
+                            } catch (RuntimeException e) {
+                                logger.log(Level.WARNING, "Failed to disconnect and delete " + c.getName() + ": " + e.getMessage());
```
barkerd427: Can we not use `comp` at this point instead of `c`?

src/main/java/org/jenkinsci/plugins/mesos/MesosCleanupThread.java
```diff
@@ -0,0 +1,110 @@
...
+                                throw e;
+                            }
+                        }
+                    });
+                    deletedNodesBuilder.add(f);
+                } else {
+                    logger.log(Level.INFO, c.getName() + " is not pending deletion");
```
barkerd427: Should this be a DEBUG as it is the normal state? This plugin already produces a lot of INFO level output.

src/main/java/org/jenkinsci/plugins/mesos/MesosCleanupThread.java
```diff
@@ -0,0 +1,110 @@
...
+                }
+            } else {
+                logger.log(Level.INFO, c.getName() + " is not a mesos computer, it is a " + c.getClass().getName());
```
barkerd427: This could also be DEBUG.

barkerd427:
This looks like a reasonable solution, and I'm guessing it's working for JClouds. I haven't tested functionality.

src/main/java/org/jenkinsci/plugins/mesos/MesosCleanupThread.java
stephenliang: Sure

src/main/java/org/jenkinsci/plugins/mesos/MesosCleanupThread.java
stephenliang: Changed it to FINE

src/main/java/org/jenkinsci/plugins/mesos/MesosCleanupThread.java
stephenliang: Changed it to FINE

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -407,6 +408,28 @@ private void createMesosTask(Offer offer, Request request) {
     LOGGER.info("Launching task " + taskId.getValue() + " with URI " +
                 joinPaths(jenkinsMaster, SLAVE_JAR_URI_SUFFIX));
 
+    for (final Computer computer : Jenkins.getInstance().getComputers()) {
```
vinodkone: I'm not sure I follow the logic here. Why are we creating a mesos task with the same name as an existing jenkins slave in the first place?

stephenliang: We are, you can see that the slave name is the name of the mesos task. It just so happens that you can be offered a task after it has been started or deleted which leads to orphan tasks and containers

On June 21, 2015 3:10:49 PM CDT, Vinod Kone <notifications@github.com> wrote:
>> @@ -407,6 +408,28 @@ private void createMesosTask(Offer offer,
>Request request) {
>>      LOGGER.info("Launching task " + taskId.getValue() + " with URI "
>+
>>                  joinPaths(jenkinsMaster, SLAVE_JAR_URI_SUFFIX));
>>  
>> +    for (final Computer computer :
>Jenkins.getInstance().getComputers()) {
>
>I'm not sure I follow the logic here. Why are we creating a mesos task
>with the same name as an existing jenkins slave in the first place? 
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/jenkinsci/mesos-plugin/pull/121/files#r32896961

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

src/main/java/org/jenkinsci/plugins/mesos/MesosRetentionStrategy.java
```diff
@@ -86,7 +86,11 @@ private long checkInternal(MesosComputer c) {
 
       if (idleMilliseconds > MINUTES.toMillis(idleTerminationMinutes)) {
         LOGGER.info("Disconnecting idle computer " + c.getName());
-        c.getNode().terminate();
+        c.getNode().setPendingDelete(true);
```
vinodkone: Why the change here to mark it for deletion instead of directly terminate it?

stephenliang: Because then you end up terminating a slave that's still doing things in the post build state, leading to a failed build. Rather we should wait until idle time has elapsed

On June 21, 2015 3:21:28 PM CDT, Vinod Kone <notifications@github.com> wrote:
>> @@ -86,7 +86,11 @@ private long checkInternal(MesosComputer c) {
>>  
>>        if (idleMilliseconds >
>MINUTES.toMillis(idleTerminationMinutes)) {
>>          LOGGER.info("Disconnecting idle computer " + c.getName());
>> -        c.getNode().terminate();
>> +        c.getNode().setPendingDelete(true);
>
>Why the change here to mark it for deletion instead of directly
>terminate it?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/jenkinsci/mesos-plugin/pull/121/files#r32897165

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

src/main/java/org/jenkinsci/plugins/mesos/MesosOneOffSlave.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Override
+    @SuppressWarnings("rawtypes")
+    public Environment setUp(AbstractBuild build, Launcher launcher, final BuildListener listener) {
+        if (MesosComputer.class.isInstance(build.getExecutor().getOwner())) {
+            final MesosComputer c = (MesosComputer) build.getExecutor().getOwner();
+            return new Environment() {
+                @Override
+                public boolean tearDown(AbstractBuild build, final BuildListener listener) throws IOException, InterruptedException {
+                    LOGGER.warning("Single-use slave " + c.getName() + " getting torn down.");
```
vinodkone: Lets stick with one terminology. either "one off slave" or "single use slave" for file names, variable names and log messages.

src/main/java/org/jenkinsci/plugins/mesos/MesosRetentionStrategy.java
vinodkone: Oh. Didn't realize that a slave is considered idle if it's doing post build things!?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
vinodkone: Sorry for not being clear. Yes, the id of the mesos task is the same as the name of the Jenkins slave and this was done to make it easy to figure out which mesos task to kill when the jenkins slave goes idle.

What I am confused about is that, this code suggests we are creating duplicate task ids which is unexpected. Can you elaborate what do you mean by your second sentence?

Thanks.

stephenliang: It's possible to receive an offer to launch the same mesos task after it has already been launched. There's nothing preventing that from happening right now and it leads to orphans. We've seen tasks that have launched in mesos but never get killed and don't do anything because they launch after it has already been launched. These if statements guard from that race condition

On June 21, 2015 3:38:04 PM CDT, Vinod Kone <notifications@github.com> wrote:
>> @@ -407,6 +408,28 @@ private void createMesosTask(Offer offer,
>Request request) {
>>      LOGGER.info("Launching task " + taskId.getValue() + " with URI "
>+
>>                  joinPaths(jenkinsMaster, SLAVE_JAR_URI_SUFFIX));
>>  
>> +    for (final Computer computer :
>Jenkins.getInstance().getComputers()) {
>
>Sorry for not being clear. Yes, the id of the mesos task is the same as
>the name of the Jenkins slave and this was done to make it easy to
>figure out which mesos task to kill when the jenkins slave goes idle.
>
>What I am confused about is that, this code suggests we are creating
>duplicate task ids which is unexpected. Can you elaborate what do you
>mean by your second sentence?
>
>Thanks.
>
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/jenkinsci/mesos-plugin/pull/121/files#r32897375

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

src/main/java/org/jenkinsci/plugins/mesos/MesosOneOffSlave.java
stephenliang: Done

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
maselvaraj: Sorry, I am also not clear on the rational behind launching the task again. I was wondering if we need to skip this task launch, if it is pending delete ?

stephenliang: Actually we should remove pending delete if it's done launching. That's in the code now, a change was made to prevent a slave from starting and then deleted mid-test

maselvaraj:
@stephenliang @vinodkone I hope all the review comments have been addressed ? 
I just have a concern on this line https://github.com/stephenliang/mesos-plugin/commit/4475bca434c0b3843534c8d5bd775c4348000d35#diff-2a5f1ead1e1896276364ab8fcf73578bR444
Can you help me understand what does it do? Trying to launch a dummy task so that it doesn't get launched again and again ?

stephenliang:
I was basically declining it. It's duplicate of declineOffer so I changed it to do that

maselvaraj:
Got it! Have you addressed all other comments ? If yes, can you please merge the latest changes from master?

stephenliang:
Okay, done

maselvaraj:
Thank you @stephenliang !

pom.xml
```diff
@@ -125,6 +126,11 @@
           <version>${assertj.version}</version>
           <scope>test</scope>
       </dependency>
+      <dependency>
+          <groupId>com.google.guava</groupId>
+          <artifactId>guava</artifactId>
+          <version>${guava.version}</version>
```
jglick: If you are not using plugin-first class loading or classpath masking, this is wrong. [Details](https://jenkins.io/doc/developer/plugin-development/dependencies-and-class-loading/)

jtnord: #322

PR #123, opened 2015-06-15
Docker Privileged Mode Support

feuer93:
Based on https://github.com/jenkinsci/mesos-plugin/pull/69 I've implemented a mergeable Version for the support of privileged mode for docker containers.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
@@ -470,7 +470,12 @@ public void setJenkinsURL(String jenkinsURL) {
 
               containerInfo = new MesosSlaveInfo.ContainerInfo(
                   containerInfoJson.getString("type"),
-                  containerInfoJson.getString("dockerImage"), volumes, parameters, networking, portMappings, containerInfoJson.getBoolean("dockerPrivilegedMode"));
+                  containerInfoJson.getString("dockerImage"),
+                  containerInfoJson.getBoolean("dockerPrivilegedMode"),
+                  volumes,
+                  parameters,
+                  networking,
+                  portMappings);
             }
 
             List<MesosSlaveInfo.URI> additionalURIs = new ArrayList<MesosSlaveInfo.URI>();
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -231,6 +236,10 @@ public String getNetworking() {
     public boolean hasPortMappings() {
       return portMappings != null && !portMappings.isEmpty();
     }
+
+    public Boolean getDockerPrivilegedMode() {
+      return dockerPrivilegedMode != null ? dockerPrivilegedMode : false;
```
vinodkone: why not just return dockerPrivilegedMode directly?

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -470,7 +470,7 @@ public boolean configure(StaplerRequest request, JSONObject object)
 
               containerInfo = new MesosSlaveInfo.ContainerInfo(
                   containerInfoJson.getString("type"),
-                  containerInfoJson.getString("dockerImage"), volumes, parameters, networking, portMappings);
+                  containerInfoJson.getString("dockerImage"), volumes, parameters, networking, portMappings, containerInfoJson.getBoolean("dockerPrivilegedMode"));
```
vinodkone: can you make the privileged mode as argument 3 to be next to "type" and "dockerImage"?

also, not yours, but mind putting each argument on its own line?

vinodkone:
Looks good. Some minor comments and I'll get this committed. Thanks!

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
feuer93: I Thougth this should be the default setting for docker containers. Nevertheless, a default setting will be set on config save, I'll change that

feuer93:
changes commited!

PR #124, opened 2015-06-16
Fix cases where you may end up with orphaned docker containers on Mesos.

stephenliang:
1) When an offer is presented to start a task that just finished
2) When the plugin kills a task and then launches another one because it was in the request queue

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -21,6 +21,7 @@ import hudson.model.Node;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -29,6 +30,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
@@ -82,6 +84,7 @@ public class JenkinsScheduler implements Scheduler {
 
   private Queue<Request> requests;
   private Map<TaskID, Result> results;
+  private Set<TaskID> finishedTasks;
   private volatile MesosSchedulerDriver driver;
   private String jenkinsMaster;
   private volatile MesosCloud mesosCloud;
@@ -92,13 +95,14 @@ public class JenkinsScheduler implements Scheduler {
   public static final Lock SUPERVISOR_LOCK = new ReentrantLock();
 
   public JenkinsScheduler(String jenkinsMaster, MesosCloud mesosCloud) {
-    LOGGER.info("JenkinsScheduler instantiated with jenkins " + jenkinsMaster +" and mesos " + mesosCloud.getMaster());
+    LOGGER.info("JenkinsScheduler instantiated with jenkins " + jenkinsMaster + " and mesos " + mesosCloud.getMaster());
 
     this.jenkinsMaster = jenkinsMaster;
     this.mesosCloud = mesosCloud;
 
     requests = new LinkedList<Request>();
     results = new HashMap<TaskID, Result>();
+    finishedTasks = Collections.newSetFromMap(new ConcurrentHashMap<TaskID, Boolean>());
   }
 
   public synchronized void init() {
@@ -213,6 +217,7 @@ public class JenkinsScheduler implements Scheduler {
         // between this removal request from jenkins and a resource getting freed up in mesos
         // resulting in scheduling the slave and resulting in orphaned task/slave not monitored
         // by Jenkins.
+
         for(Request request : requests) {
            if(request.request.slave.name.equals(name)) {
              LOGGER.info("Removing enqueued mesos task " + name);
@@ -223,8 +228,12 @@ public class JenkinsScheduler implements Scheduler {
              return;
            }
         }
+
         LOGGER.warning("Asked to kill unknown mesos task " + taskId);
     }
+
+    // Since this task is now running, we should not start this task up again at a later point in time
+    finishedTasks.add(taskId);
 
     if (mesosCloud.isOnDemandRegistration()) {
       supervise();
@@ -422,6 +431,11 @@ public class JenkinsScheduler implements Scheduler {
     LOGGER.info("Launching task " + taskId.getValue() + " with URI " +
                 joinPaths(jenkinsMaster, SLAVE_JAR_URI_SUFFIX));
 
+    if (isExistingTask(taskId)) {
+        refuseOffer(offer);
+        return;
+    }
+
     CommandInfo.Builder commandBuilder = CommandInfo.newBuilder();
     commandBuilder.setValue(
             String.format(SLAVE_COMMAND_FORMAT,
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -219,9 +226,13 @@ public synchronized void terminateJenkinsSlave(String name) {
              return;
            }
         }
+
         LOGGER.warning("Asked to kill unknown mesos task " + taskId);
     }
 
+    // Since this task is now running, we should not start this task up again at a later point in time
```
vinodkone: s/now/not/ ?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -556,6 +571,42 @@ private void createMesosTask(Offer offer, Request request) {
...
+    finishedTasks.add(taskId);
+  }
+
+  /**
+   * Checks if the given taskId already exists or just finished running. If it has, then refuse the offer.
+   * @param offer The offer to check if the task is already running
+   * @param taskId The task id
+   * @return True if the task already exists, false otherwise
+   */
+  @VisibleForTesting
```
vinodkone: why this annotation here and below? are you planning to add tests? it would be great, if you plan to by the way!

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -556,6 +571,42 @@ private void createMesosTask(Offer offer, Request request) {
...
+          LOGGER.info("Task " + taskId.getValue() + " has already finished. Ignoring and refusing offer");
+          refuseOffer(offer);
+          return true;
+      }
+
+      return false;
+  }
+
+  /**
+   * Refuses the offer provided by launcing no tasks
```
vinodkone: s/launcing/launching/

s/tasks/tasks./

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -556,6 +571,42 @@ private void createMesosTask(Offer offer, Request request) {
...
+  }
+
+  /**
+   * Checks if the given taskId already exists or just finished running. If it has, then refuse the offer.
+   * @param offer The offer to check if the task is already running
+   * @param taskId The task id
+   * @return True if the task already exists, false otherwise
+   */
+  @VisibleForTesting
+  boolean isExistingTask(Offer offer, TaskID taskId) {
```
vinodkone: the name of the function doesn't match its functionality. it's weird that a function that says whether a task exists also refuses the offer. let the function do one thing!

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
stephenliang: No, it's correct. This adds the task to the set so it won't run again later

On June 21, 2015 9:49:35 PM CDT, Vinod Kone <notifications@github.com> wrote:
>>          LOGGER.warning("Asked to kill unknown mesos task " +
>taskId);
>>      }
>>  
>> +    // Since this task is now running, we should not start this task
>up again at a later point in time
>
>s/now/not/ ?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/jenkinsci/mesos-plugin/pull/124/files#r32902602

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
stephenliang: Yep, can't add them until after a couple other pull requests are merged in though. I'll follow up with then when I get done time

On June 21, 2015 9:51:15 PM CDT, Vinod Kone <notifications@github.com> wrote:
>> @@ -556,6 +571,42 @@ private void createMesosTask(Offer offer,
>Request request) {
>>  
>>      results.put(taskId, new Result(request.result, new
>Mesos.JenkinsSlave(offer.getSlaveId()
>>          .getValue())));
>> +    finishedTasks.add(taskId);
>> +  }
>> +
>> +  /**
>> +   * Checks if the given taskId already exists or just finished
>running. If it has, then refuse the offer.
>> +   * @param offer The offer to check if the task is already running
>> +   * @param taskId The task id
>> +   * @return True if the task already exists, false otherwise
>> +   */
>> +  @VisibleForTesting
>
>why this annotation here and below? are you planning to add tests? it
>would be great, if you plan to by the way!
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/jenkinsci/mesos-plugin/pull/124/files#r32902634

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -407,6 +418,10 @@ private void createMesosTask(Offer offer, Request request) {
     LOGGER.info("Launching task " + taskId.getValue() + " with URI " +
                 joinPaths(jenkinsMaster, SLAVE_JAR_URI_SUFFIX));
 
+    if (isExistingTask(offer, taskId)) {
```
vinodkone: Do you know, when is this possible to get into this state? i.e., a createMesosTask() is called with an existing task?

stephenliang: If I remember correctly it's when you receive an multiple offers for the same task id at different times

On June 21, 2015 9:57:39 PM CDT, Vinod Kone <notifications@github.com> wrote:
>> @@ -407,6 +418,10 @@ private void createMesosTask(Offer offer,
>Request request) {
>>      LOGGER.info("Launching task " + taskId.getValue() + " with URI "
>+
>>                  joinPaths(jenkinsMaster, SLAVE_JAR_URI_SUFFIX));
>>  
>> +    if (isExistingTask(offer, taskId)) {
>
>Do you know, when is this possible to get into this state? i.e., a
>createMesosTask() is called with an existing task?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/jenkinsci/mesos-plugin/pull/124/files#r32902777

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

vinodkone: Mesos offers are not tied to tasks. Not sure what you mean by that? Did you mean createMesosTask() is being called with a duplicate 'Request'? Do you know when this happens? I'm curious because I've heard reports mentioning that the plugin tries to create duplicate tasks, but haven't been able to find the root cause yet. cc @maselvaraj

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
stephenliang: Okay, done

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
stephenliang: Moved `refuseOffer` up to the caller so that this one only does the `isExistingTask` part of the functionality.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
stephenliang: Here's one theory. On line `272`, you remove the request. However, there's no hashCode method and if anything changes inside it, then the default hashCode will change rendering it unremovable. Then when you loop through the requests again, it will get matched again and launched once more. I'm not too sure about that though, all I know is that this PR does fix the issue.

With this PR, you subvert the entire issue using the assumption that taskId is unique. On line 586, if you launch a task that is already in the list of `results`, then why should we launch it again? It will only become an orphan. On line `592`, we check the master list of tasks that are running or finished and should never run again. This prevents another class of orphaned mesos tasks.

Issue #82, opened 2015-01-29
Add ability to have a custom url for Jenkins Master used by Mesos plugins. (Use custom url and not getRootURL)


tangicolin:
Hello,
Mesos plugins use getRootURL here :https://github.com/jenkinsci/mesos-plugin/blob/master/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java#L167
If I understand well, it's needed by mesos to send back offers to Mesos Jenkins Master.  

In my setup, i have jenkins and mesos "on cloud" and i use some haproxy frontend for routing http request to jenkins Master. I want to avoid that traffic between mesos and jenkins pass thought my haproxy frontend. So for that i can set IP:PORT in jenkins url. But since 1.572, a reverse proxy check is performed by jenkins and warning is generated if jenkins url is wrong . See 
https://wiki.jenkins-ci.org/display/JENKINS/Jenkins+says+my+reverse+proxy+setup+is+broken

So i will be a nice feature if we can put a custom url into mesos jenkins plugin configuration. (And have getRootURL only as fallback if this custom url is not specified)

tomstockton:
+1 - I have exactly the same use case as @tangicolin

cornelf:
:+1:

maselvaraj:
@tangicolin @tomstockton @cornelf  I also need this as much as you do :). I tried to implement this long time back but since I faced some issues and I had to concentrate on other stuff, I did not continue later. 

It is not straight forward as it seems. Even if we place a custom URL in Mesos plugin to  download slave.jar/JNLP file.  Slave.jar would still try to connect to the root url configured in Jenkins. 

Anyways, I will work on this, will keep you posted!

maselvaraj:
@tangicolin  @tomstockton @cornelf https://github.com/jenkinsci/mesos-plugin/pull/92 - addresses this issue.

It would be great if you can build the plugin, test it and let me know how it goes.

maselvaraj:
This is implemented in v0.7.0

scottcarey:
This no longer seems to work.   
 With mesos cloud on the newest version (after update) everything breaks. 
It uses the cloud specified Jenkins URL  (call it internal-host-name) to pull the agent.jar, but then uses the global Jenkins URL (call it external-jenkins-hame) when the agent tries to register.

The logs do this successfully on a mesos node:

```
[...]

PR #131, opened 2015-06-19
Fix the JNLP using http instead of https when the jnlp URL is on http…

stephenliang:
There is a bug where if a site uses https only and forwards http to https, then when jnlp is starting it will use "http" and since it gets a 301, it throws an exception.

INFO: Locating server among [http://example.com/]
Jun 18, 2015 8:11:45 PM hudson.remoting.jnlp.Main$CuiListener error
SEVERE: http://example.com/tcpSlaveAgentListener/ is invalid: 301 MOVED PERMANENTLY
java.lang.Exception: http://example.com/tcpSlaveAgentListener/ is invalid: 301 MOVED PERMANENTLY
  at hudson.remoting.Engine.run(Engine.java:214)

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/mesos/MesosComputer/mesos-slave-agent.jnlp.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/mesos/MesosComputer/mesos-slave-agent.jnlp.jelly
@@ -45,7 +45,22 @@
         </j:if>
 
         <argument>-url</argument>
-        <argument>${app.rootUrlFromRequest}</argument>
+        <argument>${rootURL}</argument>
+
+        <j:if test="${rootURL!=app.rootUrlFromRequest}">
+          <!--
+            rootURL is based on the URL in the system config, but there has been
+            numerous reports about people moving Jenkins to another place but
+            forgetting to update it. To improve the user experience in this regard,
+            let's also pass the URL that the browser sent us as well, so that the
+            JNLP Main class can try both.
+
+            Note that rootURL is still necessary in various situations, such
+            as reverse HTTP proxy situation, which makes rootUrlFromRequest incorrect.
+          -->
+          <argument>-url</argument>
+          <argument>${app.rootUrlFromRequest}</argument>
+        </j:if>
       </application-desc>
     </jnlp>
 </j:jelly>


maselvaraj:
@stephenliang  Configuring Jenkins URL(absolute) specific to Mesos plugin didn't work for you ?

stephenliang:
No, that won't work as this jelly file uses the URL from the request, this comes in as http. In Jenkins' jelly file they have it this way, I just copied it from there

maselvaraj:
I specifically added this jelly to override default Jenkins since that causes issues when Jenkins is behind reverse proxy. 
Have you tried configuring the absolute Jenkins URL in Mesos plugin ? Can you please let me know what was the behavior if you do that ?

maselvaraj:
https://github.com/jenkinsci/mesos-plugin/pull/92

stephenliang:
No, that doesn't work. We set https on the absolute URL and it still tries to register with http. Only with this PR does it use https

maselvaraj:
This pull request falls back to the Jenkins default implementation https://github.com/kohsuke/hudson/blob/master/core/src/main/resources/hudson/slaves/SlaveComputer/slave-agent.jnlp.jelly 
The reason why I added this is to override the default behavior. If the reverse proxy server is running in a different machine, then JNLP would try to connect to that server which will fail.  Please refer https://github.com/jenkinsci/mesos-plugin/issues/82

stephenliang:
The gist of the problem is this:

Jenkins Root URL: https://example.com
Mesos Jenkins Root URL: https://example.com

JenkinsScheduler gets the https URL, it passes it to mesos-slave-agent.jnlp.jelly which uses the rootURLFromRequest. This URL uses "http", not "https", and causes the `--url` flag to be passed in as `--url http://example.com`. 

Our reverse proxy server forces all `http` traffic to `https`. This in turn responds 301 to the JNLP server and ultimately results in an exception:

   INFO: Locating server among [http://example.com/]
   Jun 18, 2015 8:11:45 PM hudson.remoting.jnlp.Main$CuiListener error
   SEVERE: http://example.com/tcpSlaveAgentListener/ is invalid: 301 MOVED PERMANENTLY
   java.lang.Exception: http://example.com/tcpSlaveAgentListener/ is invalid: 301 MOVED       PERMANENTLY
  at hudson.remoting.Engine.run(Engine.java:214)

To fix this, we are supposed to use the rootURL which uses "https" and not rootUrlFromRequest

maselvaraj:
Why would you configure reverse proxy URL in both Jenkins Root URL andMesos Jenkins Root URL ? If example.com is your reverse proxy where is Jenkins running ? 

Here is how I have configured and is working fine.
Jenkins Root URL: https://example.com 
Mesos Jenkins Root URL: http://<host where jenkins is running>:<port>

maselvaraj:
@stephenliang  Have you tried configuring Jenkins root url in Mesos plugin configuration to the actual location where your Jenkins is hosted. If it works, can you please close this pull request ?

maselvaraj:
@stephenliang Can you please close this PR? I cannot merge this since it defeats the purpose of having Jenkins root URL(absolute) specifically for Mesos plugin.

bacoboy:
Can we reopen this?  This is useful in the case where an active/standby loadbalancer is used for a HA jenkins setup.  Pointing the Mesos Plugin URL directly to the active will not switch over should the standby take over.  By putting it thru the load balancer, when the jenkins master moves, the JNLP callback also moves.  Clearly the active slaves will die, but I don't need to reconfigure jenkins when it starts at a new server.

There are also firewall exceptions at play here where the https/443 traffic is allowed, but traffic to port 8080 is blocked.

It seems more confusing to have 2 different ways to communicate with the jenkins master.  Since the JNLP auth gets added automatically, I don't see why you wouldn't use the same address mere humans use...

maselvaraj:
https://github.com/jenkinsci/mesos-plugin/issues/82

stephenliang:
But we _want_ to have traffic go through the load balancer. #82 is talking about going around it.

maselvaraj:
Let me explain it other way. 
HOST/PORT for JNLP is chosen based on the Jenkins URL. I assume you are running jenkins on machine 'A' on port 8080 and a reverse proxy like apache/nginx which listens on port 80/443 on the same box 'A' which is configured to serve Jenkins. Thats why above fix would work for you. 

Many people use reverse proxy like apache/nginx runs on a different machine other than the machine where Jenkins is hosted. In those cases this patch won't work. Slave would try to connect to the host where reverse proxy is hosted and fail.

stephenliang:
No, that's not correct. We run the reverse proxy on physical hardware seperate from the Jenkins server. The reason why we are having so much difficulty is because we force SSL protection (301 to https from http). The jelly file doesn't use the correct variable which results in "http" being used rather than "https" and Jenkins fails due to the 301, it doesn't follow the redirect.

bacoboy:
It seems that @stephenliang pull request supports both the existing mechanism (if alternate URL provided) but allows for people to use an https-only front door as well.  Can we please reconsider this?

PR #135, opened 2015-06-24
Add support for slave usage mode

### LINKED ISSUE #134: Mesos slaves are always created with Mode.NORMAL
Jenkins slaves can be created in NORMAL or EXCLUSIVE mode but the MesosSlave are always created in NORMAL mode since this value is hard coded. It would be nice the provide the ability to make that parameter configurable.

jcsirot:
This PR adds the possibility to configure the slave mode in the Slave Info section. Should Fix #134


maselvaraj:
Thanks for the PR!

PR #136, opened 2015-06-26
A value of zero indicates that the instance should never be automatically terminated

jlrigau:
This PR allows to indicate that the instance should never be automatically terminated when the value of "Idle Termination Minutes" field is equal to 0.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,9 @@
   <properties>
     <mesos.version>0.21.1</mesos.version>
     <protobuf.version>2.5.0</protobuf.version>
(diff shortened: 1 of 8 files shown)


jlrigau:
I fixed the validation for idleTerminationMinutes field and add help files.

I also fixed an issue on node configure page when showing idle information.

maselvaraj:
@jlrigau Looks good. Just couple of minor comments. Can you please address them and squash your commits into one?

maselvaraj:
Thanks for the PR!

Issue #137, opened 2015-06-26
A java.lang.NullPointerException is thrown when "Docker Privileged Mode" field is not checked


jlrigau:
This issue is linked to https://github.com/jenkinsci/mesos-plugin/pull/123

```java
SEVERE org.jenkinsci.plugins.mesos.JenkinsScheduler resourceOffers
null
java.lang.NullPointerException
	at org.jenkinsci.plugins.mesos.JenkinsScheduler.createMesosTask(JenkinsScheduler.java:495)
	at org.jenkinsci.plugins.mesos.JenkinsScheduler.resourceOffers(JenkinsScheduler.java:261)
```

maselvaraj:
I've merged your PR. Can you please close this issue?

PR #138, opened 2015-06-26
Fix java.lang.NullPointerException with privileged mode

jlrigau:
Fix issue https://github.com/jenkinsci/mesos-plugin/issues/137


maselvaraj:
Thanks for the PR!

Issue #125, opened 2015-06-17
Support new Mesos task status TASK_ERROR


geoffroyjabouley:
Currently, the Jenkins scheduler does not handle mesos TASK_ERROR status.
See https://github.com/jenkinsci/mesos-plugin/blob/master/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java#L601

This is a new status introduced in Mesos 0.21, but not used in this version (https://mesos.apache.org/blog/mesos-0-21-0-released/)
Starting from Mesos 0.22, this status can be sent.

I am currently investigating an issue where Jenkins sends a mesos slave creation request twice with the same ID, resulting in the reception of a TASK_ERROR status.
As of now, it crashes the Jenkins scheduler because of https://github.com/jenkinsci/mesos-plugin/blob/master/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java#L619

geoffroyjabouley:
Hello
any comments regarding this issue?
in the case i could do the fix, what would be the expected behavior of the plugin when receiving such TASK_ERROR mesos status.

Should we consider the same case as:
    case TASK_FAILED:
    case TASK_KILLED:
    case TASK_LOST:
      result.result.failed(result.slave);
      terminalState = true;
      break;

maselvaraj:
Hi 
Yes, it should be in the same case pointed by you.

geoffroyjabouley:
Hi

i am currently having troubles with this TASK_ERROR status.
See https://groups.google.com/forum/#!topic/jenkins-mesos/oGXfd6F7Wr8

In short:
- it seems that at some random point plugin wil send 2 requests with the same TaskID.
- The first request is received by Mesos and handled correclty --> a new slave is starting
- The second request is discarded by Mesos with the TASK_ERROR status

Now, the plugin crashes because of unmanaged TASK_ERROR status.

But i  am not sure that, in my case, the TASK_ERROR reception should terminate the slave identified by the duplicated taskID. Is there a risk that the slave started by the first request (and which is running fine btw) is terminated by using this case scenario?

Anyway, the root cause of my issue is the duplicate id request. TASK_ERROR status is just the consequence.

jeschkies:
I assume it's been fixed with #140. Feel free to reopen this issue if it's not done.

PR #140, opened 2015-06-29
Take into account new mesos status TASK_ERROR

geoffroyjabouley:
This status TASK_ERROR has been introduced in mesos 0.21, but is only sent starting from 0.22.

The PR only add the new status to the JenkinsScheduler.java in the case block of the statusUpdate() function. This avoids a Java exception crashing the scheduler.

It may need future "functional" adaptations regarding why such a status is sent by mesos.

https://github.com/jenkinsci/mesos-plugin/issues/125


maselvaraj:
@geoffroyjabouley Thanks! I've merged this so that the plugin is compatible with 0.22.X. But lets keep the issue https://github.com/jenkinsci/mesos-plugin/issues/125 open so that we shall decide and implement to handle TASK_ERROR in an appropriate way.

PR #146, opened 2015-07-06
Update config.jelly

hfufu:
Fixed missing external container image and options on loading existing configurations


maselvaraj:
There is a plan to deprecate External containerizer since it is not well supported by Mesos.
@vinodkone

PR #153, opened 2015-07-19
Cleaned up the logic of handling duplicate tasks.

vinodkone:
cc @maselvaraj @stephenliang 

Removed the 'finishedTasks' logic because all the logs I've seen so far only indicate that duplicate tasks happen when duplicate requests are enqueued. Also, removed the logic to look for pending delete slaves when looking up duplicates.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -85,7 +85,6 @@ public class JenkinsScheduler implements Scheduler {
 
   private Queue<Request> requests;
   private Map<TaskID, Result> results;
-  private Set<TaskID> finishedTasks;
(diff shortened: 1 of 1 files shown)


vinodkone:
Doesn't look the this change has anything to do with the test failure. 

testFindPortsToUseSamePortNumber test seems to be flaky on master branch as well. 
@stephenliang ?

stephenliang:
No, that's not true. It's also possible that the same task will start again after it has already started (and also when it has finished). We've observed this quite a bit but we no longer have these issues after adding `finishedTasks`.

stephenliang:
The test case checks for an infinite loop by seeing if it finishes within x milliseconds. We can reduce the flakiness by increasing that amount because an infinite loop will take infinite time to run anyway, but we can break the halting problem by using a timed stop.

vinodkone:
Abandoning this, since it's likely stale.

PR #154, opened 2015-08-03
Add jenkins label to mesos slave task's name

bacoboy:
When running many slave types, seeing everything as:

    mesos-jenkins-453c7f5c-4c70-46fa-a272-cf939852f321
    mesos-jenkins-7e2a8b6a-66e0-4cbc-87ec-cf06a4693a7e
    mesos-jenkins-06e6dd76-20e0-4b69-8fef-47d039d822f0

It becomes difficult to see what is what.

This change adds the jenkins label into the name so it looks more like this:

    mesos-foo-slave-453c7f5c-4c70-46fa-a272-cf939852f321
    mesos-bar-slave-7e2a8b6a-66e0-4cbc-87ec-cf06a4693a7e
    mesos-baz-slave-06e6dd76-20e0-4b69-8fef-47d039d822f0

where the jenkins labels are `foo-slave`, `bar-slave` and `baz-slave`

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
@@ -53,6 +53,7 @@ import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
 
 public class MesosCloud extends Cloud {
+  private static final int MAX_HOSTNAME_LENGTH = 63; // Guard against LONG hostnames - RFC-1034
   private String nativeLibraryPath;
   private String master;
   private String description;
@@ -253,7 +254,7 @@ public class MesosCloud extends Cloud {
   }
 
   private MesosSlave doProvision(int numExecutors, MesosSlaveInfo slaveInfo) throws Descriptor.FormException, IOException {
-    String name = "mesos-jenkins-" + UUID.randomUUID().toString();
+    final String name = StringUtils.abbreviate("mesos-jenkins-" + StringUtils.remove(UUID.randomUUID().toString(), '-') + "-" + slaveInfo.getLabelString(), MAX_HOSTNAME_LENGTH);
     return new MesosSlave(this, name, numExecutors, slaveInfo);
   }
 


bacoboy:
Apparently, if your label is too long, Docker won't launch it.  And by too long, it seems to be 63 characters.  A little digging shows RFC1034/1035 says each part of the hostname should be 1-63 characters.  So I'm adding a little guard code to limit this as well as change the `mesos-jenkins` prefix to just `mesos` to save characters -- since the slave label should give enough context.  And the UUID string is just to make it unique/random.

vinodkone:
The reason for the "mesos-jenkins" prefix was to clearly distinguish these slaves/tasks in the Jenkins/Mesos UIs. In Jenkins UI it is useful to have "mesos-" to know that is a cloud (mesos) slave. Similarly in the Mesos UI and logs, it makes visual sense to see "-jenkins-" in the task name, esp when the mesos cluster has multiple frameworks.

bacoboy:
swapping `jenkins` with the actual jenkins slave label tag is basically what I did.  We can add the `jenkins` back in, but since this gets used as the `hostname` parameter to the docker image, it won't launch if total length is over 63 characters.  I opted to drop some of the boilerplate (7 characters plus a dash of 63 is about 1/8th of the available space in the name) for something with more meaning.

You can make the argument that a task doesn't need the `mesos` in the name since it is obviously running in mesos when looking at the mesos console -- but clearly this is there more from the jenkins slave list side of things...

Since we have many slave types, I'm open to other suggestions on how to tell them apart if anybody has them.  It really helps to troubleshoot bad slave images to clearly see which are launching or not launching .  It also helps to track usage balance for capacity planning...

vinodkone:
Note that we can also give a 'name' to the task in addition to the 'id'.
The name is only used for debugging purposes. So one option is to add the
label info to the name and have Jenkins scheduler print the name. Of
course, it takes more effort if we also want Mesos logs to print the name.


-- Vinod

bacoboy:
My adding this Jenkins specific information to the hostname, it makes it very clear looking IN Jenkins how much is running of each type without having to cross reference the hostnames against the Mesos console.  I believe the id you are referencing appears only in the Mesos console.

By using the hostname, the linkage is clear either from looking from the Jenkins UI perspective OR from the Mesos console perspective.  It is also clear when running `docker ps` from individual Mesos slaves the origin of the container (as sometimes one is left running but isn't actually linked in Jenkins as a slave and can be killed).

brism:
:+1:

This would be very useful for us as we build out a Jenkins/Mesos CI solution with dedicated slaves per job. I will need to see this distinguishing label in both Mesos and in the Jenkins UI where it lists build slaves, so I don't think the 'name' solution would be sufficient.

kazesberger:
:+1:  we are using a very similar patch in our version of this plugin. I was about to create a pull request when i noticed there is already a better pull request. we'd very much like to have this upstream!

bacoboy:
I would love to stop running my forked version :)  How about it @vinodkone?  Seems to be enough interest.  Are there some more changes you think this patch doesn't cover?

vinodkone:
As mentioned, I would really like to keep the "jenkins" substring in the ID because it makes it easy to grep mesos logs. 

I'm worried about abbreviating UUID as it would result in (probabilistically) more ID collisions because of the reduced entropy. 

Given that we want "mesos", "jenkins" and label in the ID, I would suggest to limit the length of the label string. By my calculation it should be at most 12 chars.

bacoboy:
How about this?  I'll put the label at the end and trim there which will keep the full UUID length.  But to trim the UUID a bit, I can also remove the internal dashes since we are space constrained here...

So instead of:
`mesos-jenkins-8d643069-57b8-4e52-80c6-184e16e82b47-my-slave-label`
it would be:
`mesos-jenkins-8d64306957b84e5280c6184e16e82b47-my-slave-label`

Then I'll trim that on the end to the 63 character limit.

Better?

kazesberger:
If just the label itself gets trimmed within the computer name you could
keep it at the beginning or at least before the uuid. Or wherever want it
:-)

We also would not need to restrict label length itself.

Am 27.08.2015 16:26 schrieb "Steve Hoffman" <notifications@github.com>:
>
> How about this? I'll put the label at the end and trim there which will
keep the full UUID length. But to trim the UUID a bit, I can also remove
the internal dashes since we are space constrained here...

vinodkone:
Not sure putting 'label' at the end is enough. If label length is short (say 3 chars), then we end up trimming the UUID. Are there issues with limiting length of the label?

bacoboy:
As I said, the putting label at the end AND then trim to the 63 character limit.
Please look at the latest diff I pushed.  This is what I did PLUS removed the dashes in the uuid to keep the uniqueness requirement but provide a few more characters for use with the label (trimmed)

Vlatombe:
:+1:

Vlatombe:
To sum up the current discussions :
The current impl.
* Removes internal dashes of the UUID. That saves 3 characters
* Appends the full label (if length <= 15 chars) or trim it at the end of the task name

e.g. mesos-jenkins-453c7f5c4c7046faa272cf939852f321-mylabel
or mesos-jenkins-453c7f5c4c7046faa272cf939852f321-myveryverylongla (for label myveryverylonglabel)

The only remaining question is whether the trimmed label should be put before the uuid or after (like now)

PR #155, opened 2015-08-12
specify mesos framework role

jduan:
This attempts to address issue 68. Please take a look.

I didn't update/add any tests but the existing tests passed.

Issue #68, opened 2014-11-05
how to specify mesos framework role


veverjak:
Hi, 

I would like to use mesos roles to better weight between different frameworks. 

How do I define role for your jenkins mesos plugin?

jduan:
I'm trying to tackle this.

Vlatombe:
This has been implemented through #160

PR #160, opened 2015-09-03
Framework roles

felixb:
This is a fix for PR #155 and refs issue #68.

1. It accepts resources from `*` AND the specified role.
2. It launches tasks for offers with multiple sets of resources.

An example:

Given, you want to start a jenkins-slave with 4 cpus and 1024 mem.
Given, mesos sends an offer from a slave with:
* 2 cpus for role `jenkins`
* 3 cpus for role `*`
* 2048 mem for role `*`

The framework now will pick the 2 cpus for `jenkins`, 2 cpus for `*` and 1024 mem for `*` to start the task.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -22,6 +22,8 @@ import hudson.model.Computer;
 import hudson.model.Node;
 import hudson.util.Secret;
 
+import java.lang.Math;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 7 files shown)


felixb:
I squashed together all the commits and rebased to current master.

Vlatombe:
A couple of concerns :
* what happens if the user enters a blank role ? (I think the plugin should default to *)
* on upgrade, role will be null on existing mesos cloud instances. There should be a readResolve method setting the field to * (see https://wiki.jenkins-ci.org/display/JENKINS/Hint+on+retaining+backward+compatibility)

felixb:
I rebased to current master and made sure `*` is set as default on upgrade.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -518,29 +528,45 @@ private void detectAndAddExternalContainerInfo(Request request, CommandInfo.Buil
...
+        builder.addResources(
             Resource
                 .newBuilder()
                 .setName("mem")
                 .setType(Value.Type.SCALAR)
+                .setRole(r.getRole())
                 .setScalar(
-                    Value.Scalar
-                        .newBuilder()
-                        .setValue((1 + JVM_MEM_OVERHEAD_FACTOR) * request.request.mem)
```
Vlatombe: You removed the jvm overhead factor as part of the change. Won't that be a problem?

felixb: Thanks for the hint. I missed this change when rebaseing.
I just pushed a new version including the jvm overhead factor.

Vlatombe:
:+1:

PR #162, opened 2015-09-04
prevent offer starvation

felixb:
this change prevents offer starvation by:

* rejecting offers for longer period if no slave is queued
* calling reviveOffers when adding a slave to queue

this is necessary when running a lot of frameworks on a mesos cluster.
the rejectOfferDuration is configurable and defaults to 10min.

this is basically the same change made in marathon and chronos a few weeks ago.
see https://github.com/mesosphere/marathon/pull/1949 for further details.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -90,7 +90,7 @@ public class JenkinsScheduler implements Scheduler {
   private Queue<Request> requests;
   private Map<TaskID, Result> results;
   private Set<TaskID> finishedTasks;
-  private volatile MesosSchedulerDriver driver;
+  private volatile SchedulerDriver driver;
   private String jenkinsMaster;
   private volatile MesosCloud mesosCloud;
   private volatile boolean running;
@@ -184,6 +184,11 @@ public class JenkinsScheduler implements Scheduler {
   public synchronized void requestJenkinsSlave(Mesos.SlaveRequest request, Mesos.SlaveResult result) {
     LOGGER.info("Enqueuing jenkins slave request");
     requests.add(new Request(request, result));
+    if (driver != null) {
+      // Ask mesos to send all offers, even the those we declined earlier.
+      // See comment in resourceOffers() for further details.
+      driver.reviveOffers();
+    }
   }
 
   /**
@@ -275,6 +280,17 @@ public class JenkinsScheduler implements Scheduler {
   public synchronized void resourceOffers(SchedulerDriver driver, List<Offer> offers) {
     LOGGER.fine("Received offers " + offers.size());
     for (Offer offer : offers) {
+      if (requests.isEmpty()) {
+        // Decline offer for a longer period if no slave is waiting to get spawned.
+        // This prevents unnecessarily getting offers every few seconds and causing
+        // starvation when running a lot of frameworks.
+        double rejectOfferDuration = mesosCloud.getDeclineOfferDurationDouble();
+        LOGGER.info("No slave in queue. Rejecting offers for " + rejectOfferDuration + " ms");
+        Filters filters = Filters.newBuilder().setRefuseSeconds(rejectOfferDuration).build();
+        driver.declineOffer(offer.getId(), filters);
+        return;
+      }
+
       boolean matched = false;
       for (Request request : requests) {
         if (matches(offer, request)) {
@@ -404,6 +420,11 @@ public class JenkinsScheduler implements Scheduler {
     }
 
     return slaveTypeMatch;
+  }
+
+  @VisibleForTesting
+  void setDriver(SchedulerDriver driver) {
+    this.driver = driver;
   }
 
   @VisibleForTesting
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
@@ -73,6 +73,7 @@ import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
 
 public class MesosCloud extends Cloud {
+  private static final String DEFAULT_DECLINE_OFFER_DURATION = "600000"; // 10 mins.
   private static final int MAX_HOSTNAME_LENGTH = 63; // Guard against LONG hostnames - RFC-1034
   private String nativeLibraryPath;
   private String master;
@@ -93,6 +94,7 @@ public class MesosCloud extends Cloud {
   private final boolean checkpoint; // Set true to enable checkpointing. False by default.
   private boolean onDemandRegistration; // If set true, this framework disconnects when there are no builds in the queue and re-registers when there are.
   private String jenkinsURL;
+  private String declineOfferDuration;
 
   // Find the default values for these variables in
   // src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly.
@@ -153,7 +155,8 @@ public class MesosCloud extends Cloud {
       List<MesosSlaveInfo> slaveInfos,
       boolean checkpoint,
       boolean onDemandRegistration,
-      String jenkinsURL) throws NumberFormatException {
+      String jenkinsURL,
+      String declineOfferDuration) throws NumberFormatException {
     this("MesosCloud", nativeLibraryPath, master, description, frameworkName, slavesUser,
          principal, secret, slaveInfos, checkpoint, onDemandRegistration, jenkinsURL);
   }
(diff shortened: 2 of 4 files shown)


src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -170,6 +170,11 @@ public synchronized boolean isRunning() {
   public synchronized void requestJenkinsSlave(Mesos.SlaveRequest request, Mesos.SlaveResult result) {
     LOGGER.info("Enqueuing jenkins slave request");
     requests.add(new Request(request, result));
+    if (driver != null) {
+      // ask mesos to send all offers, even the those we declined earlier
```
vinodkone: We capitalize our sentences and end them with periods. Can you please fix here and elsewhere.

s/ask/Ask/
s/details/details./

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -261,6 +266,19 @@ public void disconnected(SchedulerDriver driver) {
   public synchronized void resourceOffers(SchedulerDriver driver, List<Offer> offers) {
     LOGGER.fine("Received offers " + offers.size());
     for (Offer offer : offers) {
+      if (requests.isEmpty()) {
+        // decline offer for a longer period if no slave is waiting to get spawned
```
vinodkone: Capitalization.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -261,6 +266,19 @@ public void disconnected(SchedulerDriver driver) {
...
+        // this prevents getting offers every few seconds
```
vinodkone: how about:

This prevents unnecessarily getting offers every few seconds and causing starvation when running a lot of frameworks.

You can kill the lines about marathon and chronos.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -53,6 +53,8 @@
 import org.kohsuke.stapler.StaplerRequest;
 
 public class MesosCloud extends Cloud {
+  private static final String DEFAULT_REJECT_OFFER_DURATION = "600000";
```
vinodkone: is this in ms? to clarify you can do 

private static final String DEFAULT_REJECT_OFFER_DURATION_MS = "600000"; // 10 mins.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -63,6 +65,7 @@
   private final boolean checkpoint; // Set true to enable checkpointing. False by default.
   private boolean onDemandRegistration; // If set true, this framework disconnects when there are no builds in the queue and re-registers when there are.
   private String jenkinsURL;
+  private String rejectOfferDuration;
```
vinodkone: s/reject/decline/

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -398,6 +403,40 @@ public void setJenkinsURL(String jenkinsURL) {
...
+  }
+
+  public double getRejectOfferDurationDouble() {
+    return Double.parseDouble(getRejectOfferDuration());
+  }
+
+  public void setRejectOfferDuration(String rejectOfferDuration) {
+    try {
+      if (rejectOfferDuration == null) {
+        LOGGER.info("Missing rejectOfferDuration. Using default " + DEFAULT_REJECT_OFFER_DURATION + ".");
```
vinodkone: please log the units of duration.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -123,7 +126,8 @@ public MesosCloud(
       List<MesosSlaveInfo> slaveInfos,
       boolean checkpoint,
       boolean onDemandRegistration,
-      String jenkinsURL) throws NumberFormatException {
+      String jenkinsURL,
+      String rejectOfferDuration) throws NumberFormatException {
```
vinodkone: s/reject/decline/

here and everywhere else in this file.

src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
```diff
@@ -45,6 +45,11 @@
                     <f:radio name="onDemandRegistration" value="false" checked="${instance.onDemandRegistration == false}" id="onDemandRegistration.false"/>
                 <st:nbsp/>${%No}
         </f:entry>
+
+        <f:entry title="${%Reject offer duration}" field="rejectOfferDuration" description="${%Decline offers for specified amount of time in case no slave is queued. The plugin asks for new offers in case a new slave needs to be started. In milliseconds.}">
```
vinodkone: s/Reject/Decline/

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -398,6 +403,40 @@ public void setJenkinsURL(String jenkinsURL) {
...
+  }
+
+  public void setRejectOfferDuration(String rejectOfferDuration) {
+    try {
+      if (rejectOfferDuration == null) {
+        LOGGER.info("Missing rejectOfferDuration. Using default " + DEFAULT_REJECT_OFFER_DURATION + ".");
+        this.rejectOfferDuration = DEFAULT_REJECT_OFFER_DURATION;
+      } else {
+        double duration = Double.parseDouble(rejectOfferDuration);
+        if (duration >= 1000) {
```
vinodkone: is it possible for the user to set a value < 1000 given the "min" condition in config.jelly?

src/test/java/org/jenkinsci/plugins/mesos/JenkinsSchedulerTest.java
```diff
@@ -213,47 +255,52 @@ public void testConstructMesosCommandInfoWithNullCustomDockerShell() throws Exce
...
-                    "test-docker-in-docker-image",
-                    Boolean.TRUE,
-                    Boolean.TRUE,
-                    useCustomDockerCommandShell,
-                    customDockerCommandShell,
-                    Collections.<MesosSlaveInfo.Volume>emptyList(),
-                    Collections.<MesosSlaveInfo.Parameter>emptyList(),
-                    Protos.ContainerInfo.DockerInfo.Network.HOST.name(),
-                    Collections.<MesosSlaveInfo.PortMapping>emptyList());
+            "docker",
```
vinodkone: why the change in indentation here?

src/test/java/org/jenkinsci/plugins/mesos/JenkinsSchedulerTest.java
```diff
@@ -213,47 +255,52 @@ public void testConstructMesosCommandInfoWithNullCustomDockerShell() throws Exce
...
+        return new Mesos.SlaveRequest(new Mesos.JenkinsSlave(TEST_JENKINS_SLAVE_NAME),0.2d,TEST_JENKINS_SLAVE_MEM,mesosSlaveInfo);
+    }
 
+    private JenkinsScheduler.Request mockMesosRequest(
+            Boolean useDocker,
+            Boolean useCustomDockerCommandShell,
+            String customDockerCommandShell) throws Descriptor.FormException {
+        Mesos.SlaveResult slaveResult = Mockito.mock(Mesos.SlaveResult.class);
+        return jenkinsScheduler.new Request(mockSlaveRequest(useDocker, useCustomDockerCommandShell,
+            customDockerCommandShell), slaveResult);
```
vinodkone: put 'mockSlaveRequest()' on next line and 'slaveResult' on the line after.

src/test/java/org/jenkinsci/plugins/mesos/JenkinsSchedulerTest.java
```diff
@@ -146,6 +149,45 @@ public void run() {
...
+
+        SchedulerDriver driver = Mockito.mock(SchedulerDriver.class);
+        Mockito.when(mesosCloud.getRejectOfferDurationDouble()).thenReturn((double) 120000);
+        jenkinsScheduler.resourceOffers(driver, offers);
+        Mockito.verify(driver).declineOffer(offer.getId());
+        Mockito.verify(driver, Mockito.never()).declineOffer(offer.getId(), Protos.Filters.newBuilder().setRefuseSeconds(120000).build());
+    }
+
+    @Test
+    public void testReviveOffersWhenAddingSlaveRequest() throws Exception {
```
vinodkone: thanks for the tests! :+1:

vinodkone:
Note that we exposed a new call in mesos called suppressOffers() that can suppress offers altogether instead of doing it one offer at a time as you did here. How about leveraging that instead?

felixb:
I worked in most of your comments.

1. I'm still declining single offers. Because the new API call was introduced in mesos 0.25, We are still running 0.24.
2. I found, that the minimum `declineOfferDuration` in the jelly config is more or less a hint only. It's still possible for a user to set values below the minimum.

felixb:
I just rebased to current master.
Any objections to merging this?

PR #168, opened 2015-09-24
Run as a user inside container (p.p. seder)

goettl79:
Needed this change to actually run our setup scripts within our docker containers as different users. We know this is a workaround, however - it served our purpose.

PS: The cmd / username used needs to be of special syntax (see commit).


goettl79:
outdated. no need to merge any more

PR #170, opened 2015-09-24
Additional fields Description and URL Format for port mappings. 

goettl79:
We use it to create a on-click webvnc url for our selenium click test builds.

p.p. seder


goettl79:
outdated, no need to merge any more

PR #171, opened 2015-09-24
Configurable additional commands to be executed on client startup

goettl79:
we use it to start a init script inside the container each time it starts. 

p.p. seder

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -632,24 +632,38 @@ public class JenkinsScheduler implements Scheduler {
         return commandBuilder;
   }
 
-  String generateJenkinsCommand2Run(int jvmMem,String jvmArgString,String jnlpArgString,String slaveName) {
+  String generateJenkinsCommand2Run(int jvmMem, String jvmArgString, String jnlpArgString, String slaveName, List<MesosSlaveInfo.Command> additionalCommands) {
 
-      return String.format(SLAVE_COMMAND_FORMAT,
-              jvmMem,
-              jvmArgString,
-              jnlpArgString,
-              getJnlpSecret(slaveName),
-              getJnlpUrl(slaveName));
+    String slaveCmd = String.format(SLAVE_COMMAND_FORMAT,
+            jvmMem,
+            jvmArgString,
+            jnlpArgString,
+            getJnlpSecret(slaveName),
+            getJnlpUrl(slaveName));
+
+    StringBuilder commandStringBuilder = new StringBuilder();
+
+    if (additionalCommands != null && !additionalCommands.isEmpty()) {
+      for (MesosSlaveInfo.Command additionalCommand : additionalCommands) {
+        commandStringBuilder.append(additionalCommand.getValue() + " && ");
+      }
+      commandStringBuilder.append("exec ");
+      commandStringBuilder.append(slaveCmd);
+      return commandStringBuilder.toString();
+    }
+
+    return slaveCmd;
   }
 
   private CommandInfo.Builder getBaseCommandBuilder(Request request) {
 
         CommandInfo.Builder commandBuilder = CommandInfo.newBuilder();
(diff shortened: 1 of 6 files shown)


src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
```diff
@@ -238,6 +238,24 @@
                                 </fieldset>
                             </f:repeatable>
                         </f:entry>
+
+                        <f:entry title="${%Additional Commands}" field="additionalCommands">
```
vinodkone: can you add a help file for this?

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -357,6 +364,20 @@ public boolean isReadOnly() {
...
+
+      @DataBoundConstructor
+      public Command(String value) {
+        this.value = value;
+      }
+
+      public String getValue() {
+        return value;
+      }
+
```
vinodkone: kill new line.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -513,6 +516,18 @@ public boolean configure(StaplerRequest request, JSONObject object)
...
             }
+
+
+            List<MesosSlaveInfo.Command> additionalCommands = new ArrayList<MesosSlaveInfo.Command>();
+            if (label.has("additionalCommands")) {
+              JSONArray additionalCommandsJson = label.getJSONArray("additionalCommands");
+              for (Object obj : additionalCommandsJson) {
+                JSONObject additionalCommandJson = (JSONObject) obj;
+                additionalCommands.add(new MesosSlaveInfo.Command(
+                        additionalCommandJson.getString("value")));
```
vinodkone: indent by 4 spaces.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -52,6 +52,9 @@
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
 
+
+import java.util.*;
```
vinodkone: can you just import what's needed?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -632,24 +632,38 @@ private void getContainerInfoBuilder(Offer offer, Request request, String slaveN
...
+
+    return slaveCmd;
   }
 
   private CommandInfo.Builder getBaseCommandBuilder(Request request) {
 
         CommandInfo.Builder commandBuilder = CommandInfo.newBuilder();
         String jenkinsCommand2Run = generateJenkinsCommand2Run(
-            request.request.mem,
+            request.request.slaveInfo.getSlaveMem(),
```
vinodkone: does this include the memory needed for the executors?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -632,24 +632,38 @@ private void getContainerInfoBuilder(Offer offer, Request request, String slaveN
...
+            jvmArgString,
+            jnlpArgString,
+            getJnlpSecret(slaveName),
+            getJnlpUrl(slaveName));
+
+    StringBuilder commandStringBuilder = new StringBuilder();
+
+    if (additionalCommands != null && !additionalCommands.isEmpty()) {
+      for (MesosSlaveInfo.Command additionalCommand : additionalCommands) {
+        commandStringBuilder.append(additionalCommand.getValue() + " && ");
```
vinodkone: hmm. i think it won't be obvious to users that the commands are tied by "&&". how about having user input just one shell command instead of a list? the user can easily create a composite shell command (using && or ||) himself.

goettl79:
Also here, no need to merge this any more. seems to be outdated

Issue #164, opened 2015-09-22
Termination of many nodes


digitalrinaldo:
After the jenkins mesos plugin has been given a resource to use and the use of the resource is complete the Jenkins UI is unreadable, nodes are left in the following state for 1 minute.  Can this time be reduced or can the node be removed from the ui in some way.
<img width="323" alt="screen shot 2015-09-22 at 5 14 59 pm" src="https://cloud.githubusercontent.com/assets/799458/10032035/8c1b3668-614e-11e5-9063-6c87b9c88898.png">

vinodkone:
These Jenkins slaves should go offline (and removed from the Jenkins UI)
when they are idle for a configured amount of 'idle termination' time
(default might be 1 min). I'm a bit surprised that offline slaves stay for
1 minute in the UI. It's probably because of a recent code change that
introduced an asynchronous cleanup thread.


-- Vinod

jeschkies:
Closed by #173.

PR #173, opened 2015-09-27
Implement Maximum Time To Live

goettl79:


digitalrinaldo:
This looks like it will solve a problem we have been having in #164 I would like to build and try to understand your work in this area.  I am not familiar with the pull and merge system being used here. How can I get a source copy of the mesos plugin to build that has your change.  Are you interested in feedback? Thanks for all the work you did, reading your code helped me understand how some parts of the plugin work.

goettl79:
Always interested in feeback :-). 

git clone https://github.com/goettl79/mesos-plugin.git
git checkout  maximum_ttl_pull
mvn install

 -> you will get a new version of the plugin, however I don't think this will fix your problem. However, The issue you have ( #164 ) is more likely a jenkins issue.

Regards
georg

src/main/java/org/jenkinsci/plugins/mesos/MesosRetentionStrategy.java
```diff
@@ -94,6 +96,22 @@ private long checkInternal(MesosComputer c) {
...
+    // Terminate the computer if it is exists for longer than
+    // 'maximumTimeToLive'.
+    final long timeLivedInMilliseconds =
+            DateTimeUtils.currentTimeMillis() - c.getConnectTime();
+
+    if (c.isOnline()) {
+
+      if (timeLivedInMilliseconds > MINUTES.toMillis(maximumTimeToLive)) {
+        LOGGER.info("Disconnecting computer greater maximum TTL " + c.getName());
+
```
ryoichitaniguchi: Let me ask a question. Can slave be removed without setting pendingDelete, is anything else of safe clener here except for MesosCleanupThread ?
(I look into these piece of  codes recently, suffering from slave connectivety issue between plugin and mesos/zookeeper in my test environment.)

goettl79: I don't think there is something better than setting a pending delete. This
will give jenkins the time to clean up.
Am 14.06.2016 03:27 schrieb "Ryoichi Taniguchi" <notifications@github.com>:

goettl79:
I think this won't be merged... closing it

PR #174, opened 2015-10-01
[JENKINS-30741] - Support the creation on NonModifyableCloudConfigs

oleg-nenashev:
On a multi-instanceJenkins infrastructure we sometimes need to provide an alternate implementation of MesosCloud, which would hide the most of configs from local administrators.

It would be great to have a generic MesosCloud implementation, which takes a configuration and then hides parameters from users and prevents their configuration from UIs.

https://issues.jenkins-ci.org/browse/JENKINS-30741

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
@@ -38,6 +38,7 @@ import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import javax.annotation.Nonnull;
 
 import javax.servlet.ServletException;
 
@@ -124,7 +125,29 @@ public class MesosCloud extends Cloud {
       boolean checkpoint,
       boolean onDemandRegistration,
       String jenkinsURL) throws NumberFormatException {
-    super("MesosCloud");
+    this("MesosCloud", nativeLibraryPath, master, description, frameworkName, slavesUser, 
+            principal, secret, slaveInfos, checkpoint, onDemandRegistration, jenkinsURL);
+  }
+  
+  /**
+   * Constructor, which also allows to specify a custom name.
+   * @throws NumberFormatException Numeric parameter parsing error
+   * @since TODO
+   */
+  protected MesosCloud(
+      String cloudName,
+      String nativeLibraryPath,
+      String master,
+      String description,
+      String frameworkName,
+      String slavesUser,
+      String principal,
+      String secret,
+      List<MesosSlaveInfo> slaveInfos,
+      boolean checkpoint,
+      boolean onDemandRegistration,
+      String jenkinsURL) throws NumberFormatException {
+    super(cloudName);
 
     this.nativeLibraryPath = nativeLibraryPath;
     this.master = master;
@@ -147,6 +170,20 @@ public class MesosCloud extends Cloud {
     }
   }
 
+  /**
+   * Copy constructor.
+   * Allows to create copies of the original mesos cloud. Since it's a singleton
+   * by design, this method also allows specifying a new name.
+   * @param name Name of the cloud to be created
+   * @param source Source Mesos cloud implementation
+   * @since TODO
+   */
+  public MesosCloud(@Nonnull String name, @Nonnull MesosCloud source) {
+      this(name, source.nativeLibraryPath, source.master, source.description, source.frameworkName,
+           source.slavesUser, source.principal, source.secret, source.slaveInfos, source.checkpoint,
+           source.onDemandRegistration, source.jenkinsURL);
+  }
+          
   // Since MesosCloud is used as a key to a Hashmap, we need to set equals/hashcode
   // or lookups won't work if any fields are changed.  Use master string as the key since
   // the rest of this code assumes it is unique among the Cloud objects.
--- a/src/main/java/org/jenkinsci/plugins/mesos/NonConfigurableMesosCloud.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/NonConfigurableMesosCloud.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2015 CloudBees, Inc. and other contributors.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jenkinsci.plugins.mesos;
+
+import hudson.Extension;
+import hudson.slaves.Cloud;
+import java.util.List;
+import jenkins.model.Jenkins;
+import net.sf.json.JSONObject;
+import org.kohsuke.stapler.StaplerRequest;
+
+/**
+ * Mesos {@link Cloud}, which cannot be reconfigured from the web interface.
+ * @author Oleg Nenashev
+ * @since TODO
+ */
+public class NonConfigurableMesosCloud extends MesosCloud {
+ 
+    /**
+     * Constructs a cloud from a parent Mesos Cloud
+     * @param name Name of the cloud to be created
+     */
+    public NonConfigurableMesosCloud(String name, String nativeLibraryPath, String master, String description, 
+            String frameworkName, String slavesUser, String principal, String secret, 
+            List<MesosSlaveInfo> slaveInfos, boolean checkpoint, boolean onDemandRegistration, String jenkinsURL) 
+            throws NumberFormatException {
+        super(name, nativeLibraryPath, master, description, frameworkName, slavesUser, 
+                principal, secret, slaveInfos, checkpoint, onDemandRegistration, jenkinsURL);
+    }
+    
+    /**
+     * Copy constructor.
+     * @param name Name of the cloud to be created.
+     * @param cloud Cloud to be copied
+     */
+    public NonConfigurableMesosCloud(String name, MesosCloud cloud) {
+        super(name, cloud);
+    }
+    
+    @Extension
+    public static class DescriptorImpl extends MesosCloud.DescriptorImpl {
+        @Override
+        public String getDisplayName() {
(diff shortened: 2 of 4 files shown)


src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -124,7 +125,29 @@ public MesosCloud(
...
+            principal, secret, slaveInfos, checkpoint, onDemandRegistration, jenkinsURL);
+  }
+  
+  /**
+   * Constructor, which also allows to specify a custom name.
+   * @throws NumberFormatException Numeric parameter parsing error
+   * @since TODO
+   */
+  protected MesosCloud(
+      String cloudName,
```
oleg-nenashev: TODO: fix autoformatting

oleg-nenashev:
Removed the outdated screenshot

oleg-nenashev:
To reviewers: <code> NonConfigurableMesosCloud</code> is optional and may be removed from this PR upon the feedback. I mostly need a copy constructor, which would allow to implement similar class in an external plugin.

Thanks in advance

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -124,7 +125,29 @@ public MesosCloud(
       boolean checkpoint,
       boolean onDemandRegistration,
       String jenkinsURL) throws NumberFormatException {
-    super("MesosCloud");
+    this("MesosCloud", nativeLibraryPath, master, description, frameworkName, slavesUser,
```
lordofthejars: The only thing I see here is that this constructor has a lot of parameters and of similar type `String`. It is not bad but it can be easy for a developer that wants to use this class to misplaced some of the parameters during the instantiation. Maybe it could be interesting to create some kind of wrapper classes that groups the parameters for meaning, for example `Security`, `Slaves`, ...

oleg-nenashev: I would +1 one for it, but it is out of the scope of this pull request

lordofthejars: Ok. I have seen the changes and the constructor (this(....)) was modified too, for this reason I thought that this issue had introduced this.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -147,6 +170,20 @@ public MesosCloud(
...
+   * by design, this method also allows specifying a new name.
+   * @param name Name of the cloud to be created
+   * @param source Source Mesos cloud implementation
+   * @since TODO
+   */
+  public MesosCloud(@Nonnull String name, @Nonnull MesosCloud source) {
+      this(name, source.nativeLibraryPath, source.master, source.description, source.frameworkName,
+           source.slavesUser, source.principal, source.secret, source.slaveInfos, source.checkpoint,
+           source.onDemandRegistration, source.jenkinsURL);
+  }
```
tfennelly: How about putting this into a `clone` method?

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -124,7 +125,29 @@ public MesosCloud(
...
+      String description,
+      String frameworkName,
+      String slavesUser,
+      String principal,
+      String secret,
+      List<MesosSlaveInfo> slaveInfos,
+      boolean checkpoint,
+      boolean onDemandRegistration,
+      String jenkinsURL) throws NumberFormatException {
+    super(cloudName);
```
tfennelly: I know you're adding to existing code (and so the pattern has already been set) but it seems like a fluent API on this class would be nicer than these constructors with lots of (what appear to be - no checks on them) optional args.

varmenise:
don t have strong opinion on the arguments, but :bee: for me

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
oleg-nenashev: It's not a clone() operation, because we change the name.
Clone() call also presumes that we will clone all fields beneath, but we don't do it here.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
oleg-nenashev: +1, but outside of the scope

jglick:
BTW I am not sure whether `NonModifyableCloudConfigs` is intended to be a class name, but if so, it is misspelled (should be _modifiable_).

oleg-nenashev:
@reviewbybees done

No sense to fix typos before we get a response from the plugin owner

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -124,7 +125,29 @@ public MesosCloud(
...
       String jenkinsURL) throws NumberFormatException {
-    super("MesosCloud");
+    this("MesosCloud", nativeLibraryPath, master, description, frameworkName, slavesUser, 
+            principal, secret, slaveInfos, checkpoint, onDemandRegistration, jenkinsURL);
+  }
+  
+  /**
+   * Constructor, which also allows to specify a custom name.
+   * @throws NumberFormatException Numeric parameter parsing error
+   * @since TODO
```
vinodkone: Incomplete sentence?

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -124,7 +125,29 @@ public MesosCloud(
       boolean checkpoint,
       boolean onDemandRegistration,
       String jenkinsURL) throws NumberFormatException {
-    super("MesosCloud");
+    this("MesosCloud", nativeLibraryPath, master, description, frameworkName, slavesUser, 
+            principal, secret, slaveInfos, checkpoint, onDemandRegistration, jenkinsURL);
```
vinodkone: please indent 'principal' with "MesosCloud".

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -147,6 +170,20 @@ public MesosCloud(
     }
   }
 
+  /**
+   * Copy constructor.
+   * Allows to create copies of the original mesos cloud. Since it's a singleton
+   * by design, this method also allows specifying a new name.
+   * @param name Name of the cloud to be created
+   * @param source Source Mesos cloud implementation
+   * @since TODO
```
vinodkone: incomplete?

src/main/java/org/jenkinsci/plugins/mesos/NonConfigurableMesosCloud.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        @Override
+        public boolean configure(StaplerRequest request, JSONObject object) throws FormException {
+            return true;
+        }
+        
+        @Override
+        public Cloud newInstance(StaplerRequest req, JSONObject formData) throws FormException {
+            // TODO: replace by getActieInstance in 1.590+
```
vinodkone: s/getActie/getActive/ ?

src/main/java/org/jenkinsci/plugins/mesos/NonConfigurableMesosCloud.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Constructs a cloud from a parent Mesos Cloud
+     * @param name Name of the cloud to be created
+     */
+    public NonConfigurableMesosCloud(String name, String nativeLibraryPath, String master, String description, 
+            String frameworkName, String slavesUser, String principal, String secret, 
+            List<MesosSlaveInfo> slaveInfos, boolean checkpoint, boolean onDemandRegistration, String jenkinsURL) 
+            throws NumberFormatException {
+        super(name, nativeLibraryPath, master, description, frameworkName, slavesUser, 
+                principal, secret, slaveInfos, checkpoint, onDemandRegistration, jenkinsURL);
```
vinodkone: indent 'principal' with 'name'

src/main/java/org/jenkinsci/plugins/mesos/NonConfigurableMesosCloud.java
```diff
@@ -0,0 +1,77 @@
...
+import jenkins.model.Jenkins;
+import net.sf.json.JSONObject;
+import org.kohsuke.stapler.StaplerRequest;
+
+/**
+ * Mesos {@link Cloud}, which cannot be reconfigured from the web interface.
+ * @author Oleg Nenashev
+ * @since TODO
+ */
+public class NonConfigurableMesosCloud extends MesosCloud {
```
vinodkone: What's the UX with the NonConfigurableMesosCloud? Can you show me some screen shots? Is it the case that, once this type of cloud is created, it cannot be re-configured but it can be copied?

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
oleg-nenashev: Which one? TODO means that there should be a version when the PR gets released

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
oleg-nenashev: ok

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
oleg-nenashev: see above. "since" Javadoc field requires a version, but it can be specified only when/if the PR gets merged

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
vinodkone: I see. Just set the version to the current version in the POM please. 0.9.0.

src/main/java/org/jenkinsci/plugins/mesos/NonConfigurableMesosCloud.java
vinodkone: I would propose to remove the NonConfigurableMesosClould class from this PR to cut the scope. Sounds like it is not really needed for your plugin to work.

Vlatombe:
Merged as https://github.com/jenkinsci/mesos-plugin/commit/0b32fd5f47c051c77b9e7f02bf6904dbf3cecaf7 . I removed the NonConfigurableMesosCloud class.

Issue #176, opened 2015-10-14
Cannot add a MesosSlave node


kaysoky:
Plugin version: 2461ae1fdf67730201e1df3643a5aa94aa5906d8
Jenkins version: 1.565.3

## Problem
Note: The textboxes on `/jenkins/computer/createItem` are read-only.  So I had to modify the DOM to get past this form.  Or, you can modify `src/main/resources/org/jenkinsci/plugins/mesos/MesosSlave/configure-entries.jelly` with `s/readOnlyTextbox/textbox/g`.

When I try to add a MesosSlave node (via `/jenkins/computer/new`), upon clicking "save" (`/jenkins/computer/doCreateItem`) I run into this error:
```
javax.servlet.ServletException: java.lang.RuntimeException: Failed to instantiate class org.jenkinsci.plugins.mesos.MesosSlave from {"name":"asdf","numExecutors":"1","labelString":"","idleTerminationMinutes":"3","type":"org.jenkinsci.plugins.mesos.MesosSlave"}
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:783)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:863)
	at org.kohsuke.stapler.MetaClass$4.doDispatch(MetaClass.java:210)
	at org.kohsuke.stapler.NameBasedDispatcher.dispatch(NameBasedDispatcher.java:53)
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:733)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:863)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:636)
	at org.kohsuke.stapler.Stapler.service(Stapler.java:225)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:820)
	at org.mortbay.jetty.servlet.ServletHolder.handle(ServletHolder.java:491)
	at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1074)
	at hudson.util.PluginServletFilter$1.doFilter(PluginServletFilter.java:96)
	at hudson.util.PluginServletFilter.doFilter(PluginServletFilter.java:88)
	at org.mortbay.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1065)
	at hudson.security.csrf.CrumbFilter.doFilter(CrumbFilter.java:48)
[...]

PR #177, opened 2015-10-14
Quality of life improvements for newbies.

kaysoky:
Add text to dissuade usage of the /computer/new endpoint.


kaysoky:
These are changes to prevent another person (like myself) from bumbling into some known issues, like #176.

vinodkone:
Does a blank jenkinsURL really throw a NPE? i'm surprised because that didn't use to be the case. can you paste the trace?

kaysoky:
Looks like I was doing something wrong before.  I can't repro the NPE anymore :(

PR #178, opened 2015-10-20
Allow running all jobs in mesos slaves

carlossg:
If label is empty then mesos slaves will be used for all jobs without label requirements

In a mesos-only Jenkins installation it is confusing for users to create jobs
that sit there because the label is by default empty and never get scheduled in Mesos.
If the Jenkins config for mesos cloud has an empty label then the mesos slaves
are added to the default pool for any job without label requirements

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
@@ -272,9 +272,9 @@ public class MesosCloud extends Cloud {
     // item in the list of configured Mesos labels.
     // TODO(vinod): The framework may not have the resources necessary
     // to start a task when it comes time to launch the slave.
-    if (label != null && slaveInfos != null) {
+    if (slaveInfos != null) {
       for (MesosSlaveInfo slaveInfo : slaveInfos) {
-        if (label.matches(Label.parse(slaveInfo.getLabelString()))) {
+        if (slaveInfo.matchesLabel(label)) {
           return true;
         }
       }
@@ -365,7 +365,7 @@ public class MesosCloud extends Cloud {
   private MesosSlaveInfo getSlaveInfo(List<MesosSlaveInfo> slaveInfos,
       Label label) {
     for (MesosSlaveInfo slaveInfo : slaveInfos) {
-      if (label.matches(Label.parse(slaveInfo.getLabelString()))) {
+      if (slaveInfo.matchesLabel(label)) {
         return slaveInfo;
       }
     }
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
@@ -169,10 +169,10 @@ public class MesosSlaveInfo {
   }
 
   /**
-   * Check if the label in the slave matches the provided label, either both are null or are the same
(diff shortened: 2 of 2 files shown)


stephenc:
:bee:

oleg-nenashev:
@reviewbybees done

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -4,11 +4,15 @@
...
+import java.util.Set;
 import java.util.logging.Logger;
 
+import javax.annotation.CheckForNull;
+
+import hudson.model.Label;
 import hudson.model.Node;
 import hudson.model.Node.Mode;
-
+import hudson.model.labels.LabelAtom;
```
vinodkone: unused?

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -272,9 +272,9 @@ public boolean canProvision(Label label) {
     // item in the list of configured Mesos labels.
     // TODO(vinod): The framework may not have the resources necessary
     // to start a task when it comes time to launch the slave.
-    if (label != null && slaveInfos != null) {
+    if (slaveInfos != null) {
       for (MesosSlaveInfo slaveInfo : slaveInfos) {
-        if (label.matches(Label.parse(slaveInfo.getLabelString()))) {
+        if (slaveInfo.matchesLabel(label)) {
```
vinodkone: Just to confirm, does this still work in a non-mesos only jenkins installation? i.e., if all of the mesos slaves are configured with a non-null label, do jobs without labels run on jenkins master or non-mesos jenkins slaves?

carlossg: yes, this PR doesn't change the default behavior, if job doesn't have a label it will only run in slaves that don't have a label. By default mesos slaves will still use the `mesos` label

carlossg:
anything else I can do to get this merged?

tfennelly:
:bee: lgtm

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -166,6 +168,17 @@ private String cleanseJvmArgs(final String jvmArgs) {
     return jvmArgs.replaceAll(JVM_ARGS_PATTERN, "");
   }
 
+  /**
+   * Check if the label in the slave matches the provided label, either both are null or are the same
```
vinodkone: period at the end of the sentence.

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -166,6 +168,17 @@ private String cleanseJvmArgs(final String jvmArgs) {
...
+   * 
+   * @param label
+   * @return whether the slave label matches
```
vinodkone: s/whether/Whether/
s/matches/matches./

vinodkone:
looking good. some minor fixes for documentation. i'll commit them as soon as you fix them. thanks.

carlossg:
done

carlossg:
can we get a new release?

vinodkone:
Happy to cut a new release. Will start a new thread on the mailing list.

PR #179, opened 2015-10-26
Integration with credentials plugin

Vlatombe:
This pull request provides integration with credentials plugin to replace framework principal/secret.

Credentials plugin provides a central facility to manage credentials and to update them only once if needed.

The motivation for this change is to provide the ability to update programmatically the mesos credentials without touching the cloud configuration, which may have been altered by an administrator. The current implementation doesn't allow that in an easy way, as you need to post the whole configuration.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -16,9 +16,11 @@
 package org.jenkinsci.plugins.mesos;
 
 
+import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
 import com.google.common.annotations.VisibleForTesting;
 import hudson.model.Computer;
 import hudson.model.Node;
+import hudson.util.Secret;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -118,25 +120,28 @@ public class JenkinsScheduler implements Scheduler {
           String targetUser = mesosCloud.getSlavesUser();
           String webUrl = Jenkins.getInstance().getRootUrl();
           if (webUrl == null) webUrl = System.getenv("JENKINS_URL");
+          StandardUsernamePasswordCredentials credentials = mesosCloud.getCredentials();
+          String principal = credentials == null ? "jenkins" : credentials.getUsername();
+          String secret = credentials == null ? "" : Secret.toString(credentials.getPassword());
           // Have Mesos fill in the current user.
           FrameworkInfo framework = FrameworkInfo.newBuilder()
             .setUser(targetUser == null ? "" : targetUser)
             .setName(mesosCloud.getFrameworkName())
-            .setPrincipal(mesosCloud.getPrincipal())
+            .setPrincipal(principal)
             .setCheckpoint(mesosCloud.isCheckpoint())
             .setWebuiUrl(webUrl != null ? webUrl : "")
             .build();
 
           LOGGER.info("Initializing the Mesos driver with options"
             + "\n" + "Framework Name: " + framework.getName()
-            + "\n" + "Principal: " + framework.getPrincipal()
+            + "\n" + "Principal: " + principal
             + "\n" + "Checkpointing: " + framework.getCheckpoint()
           );
 
-          if (StringUtils.isNotBlank(mesosCloud.getSecret())) {
+          if (StringUtils.isNotBlank(secret)) {
             Credential credential = Credential.newBuilder()
-              .setPrincipal(mesosCloud.getPrincipal())
-              .setSecret(ByteString.copyFromUtf8(mesosCloud.getSecret()))
+              .setPrincipal(principal)
+              .setSecret(ByteString.copyFromUtf8(secret))
               .build();
 
             LOGGER.info("Authenticating with Mesos master with principal " + credential.getPrincipal());
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
@@ -14,15 +14,28 @@
  */
 package org.jenkinsci.plugins.mesos;
 
+import com.cloudbees.plugins.credentials.CredentialsMatchers;
+import com.cloudbees.plugins.credentials.CredentialsProvider;
+import com.cloudbees.plugins.credentials.CredentialsScope;
+import com.cloudbees.plugins.credentials.SystemCredentialsProvider;
+import com.cloudbees.plugins.credentials.common.StandardListBoxModel;
+import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
+import com.cloudbees.plugins.credentials.common.UsernamePasswordCredentials;
+import com.cloudbees.plugins.credentials.domains.DomainRequirement;
+import com.cloudbees.plugins.credentials.domains.URIRequirementBuilder;
+import com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl;
+
 import hudson.Extension;
 import hudson.init.InitMilestone;
 import hudson.init.Initializer;
 import hudson.model.Computer;
 import hudson.model.Descriptor;
 import hudson.model.Hudson;
+import hudson.model.Item;
 import hudson.model.Label;
 import hudson.model.Node;
 import hudson.model.Node.Mode;
+import hudson.security.ACL;
 import hudson.slaves.Cloud;
 import hudson.slaves.NodeProvisioner.PlannedNode;
 import hudson.util.FormValidation;
@@ -41,6 +54,8 @@ import java.util.logging.Logger;
 
 import javax.servlet.ServletException;
 
+import hudson.util.ListBoxModel;
+import hudson.util.Secret;
 import jenkins.model.Jenkins;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
@@ -48,7 +63,11 @@ import net.sf.json.JSONObject;
 import org.apache.commons.lang.StringUtils;
 import org.apache.mesos.MesosNativeLibrary;
 import org.apache.mesos.Protos.ContainerInfo.DockerInfo.Network;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.DoNotUse;
+import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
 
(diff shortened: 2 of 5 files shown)


Vlatombe:
@reviewbybees

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -417,6 +504,17 @@ public String getDisplayName() {
...
+    @SuppressWarnings("unused") // by stapler
+    public ListBoxModel doFillCredentialsIdItems(@AncestorInPath Item item, @QueryParameter String master) {
+      List<DomainRequirement> domainRequirements = (master == null) ? Collections.<DomainRequirement>emptyList()
+              : URIRequirementBuilder.fromUri(master.trim()).build();
+      return new StandardListBoxModel().withEmptySelection().withMatching(
+              CredentialsMatchers.instanceOf(UsernamePasswordCredentials.class),
+              CredentialsProvider.lookupCredentials(StandardUsernamePasswordCredentials.class, item, null,
+                      domainRequirements));
+    }
+
```
recena: Are you using 4 spaces or 2 spaces? It is mixed in this plugin.

Vlatombe: Should be 2 spaces, I fixed occurrences in MesosCloud.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -58,8 +76,11 @@
   private String description;
   private String frameworkName;
   private String slavesUser;
-  private String principal;
-  private String secret;
+  private String credentialsId;
+  @Deprecated
+  private transient String principal;
+  @Deprecated
+  private transient String secret;
```
tfennelly: Looking at the `readResolve` method, it seems `principal` and `secret` get reset to `null` after being used to resolve a `credentialsId`. Is there a need to track calls to the setters methods for these properties and then run the equivalent of the `readResolve` again? I assume the fact that the setter methods exist may mean someone can call them.

Vlatombe: I don't think I want to track these.
* I don't think the setters are being used at the moment
* I'd rather change setters implementations to throw Exceptions to prevent anyone using them, or even delete the methods to force potential client to use the credentials management instead.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
recena: @Vlatombe Great!

Vlatombe:
@vinodkone ping

stephenc:
:bee:

tfennelly:
:bee: 

Assuming the build failure is expected/flaky.

Vlatombe:
@reviewbybees done

src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
```diff
@@ -19,12 +19,8 @@
         <f:textbox field="slavesUser" default=""/>
     </f:entry>
 
-    <f:entry title="${%Framework Principal}" field="principal">
-      <f:textbox field="principal" default="jenkins" />
-    </f:entry>
-
-    <f:entry title="${%Framework Secret}" field="secret" description="${%Specify secret to enable framework authentication}">
-      <f:textbox field="secret"/>
+    <f:entry field="credentialsId" title="${%Framework credentials}">
```
vinodkone: Is this a backwards compatible change? If not, we can just remove these fields.

src/test/java/org/jenkinsci/plugins/mesos/MesosCloudTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import static org.junit.Assert.assertTrue;
+
+public class MesosCloudTest {
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    /**
+     * Ignored because I'm getting Cannot call method "hasClassName" of undefined (http://localhost:55154/jenkins/adjuncts/b4f04185/lib/form/hetero-list/hetero-list.js#16)
```
vinodkone: Please fix this test. Adding a test that doesn't work sets a bad precedent.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -417,6 +507,17 @@ public String getDisplayName() {
       return "Mesos Cloud";
     }
 
+    @Restricted(DoNotUse.class) // stapler only
```
vinodkone: We use fully formed sentences for our comments, i.e., we use proper capitalization and periods.

// Stapler only.

Please fix and everywhere else in this patch.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -390,6 +453,33 @@ public JSONObject getSlaveAttributeForLabel(String labelName) {
     return null;
   }
 
+  protected Object readResolve() {
```
vinodkone: It's really hard to understand what's happening in this method. Even the name of the method doesn't give me a strong clue. This could use some comments.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -208,6 +229,16 @@ public void restartMesos() {
 
   }
 
+  public StandardUsernamePasswordCredentials getCredentials() {
```
vinodkone: Add some comments please.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -58,8 +76,11 @@
   private String description;
   private String frameworkName;
   private String slavesUser;
-  private String principal;
-  private String secret;
+  private String credentialsId;
+  @Deprecated
```
vinodkone: We cannot deprecate these without giving a heads up to existing users. Can you please email our mailing list and get feedback about this change?

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -14,15 +14,27 @@
  */
 package org.jenkinsci.plugins.mesos;
 
+import com.cloudbees.plugins.credentials.CredentialsMatchers;
```
vinodkone: I'm assuming these packages are all open source without any licensing conflicts?

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -14,15 +14,27 @@
...
+import com.cloudbees.plugins.credentials.CredentialsProvider;
+import com.cloudbees.plugins.credentials.CredentialsScope;
+import com.cloudbees.plugins.credentials.SystemCredentialsProvider;
+import com.cloudbees.plugins.credentials.common.StandardListBoxModel;
+import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
+import com.cloudbees.plugins.credentials.common.UsernamePasswordCredentials;
+import com.cloudbees.plugins.credentials.domains.DomainRequirement;
+import com.cloudbees.plugins.credentials.domains.URIRequirementBuilder;
+import com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl;
 import hudson.Extension;
```
vinodkone: new line.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -114,42 +116,49 @@ public synchronized void init() {
...
             driver = new MesosSchedulerDriver(JenkinsScheduler.this, framework, mesosCloud.getMaster(), credential);
-        } else {
+          } else {
             driver = new MesosSchedulerDriver(JenkinsScheduler.this, framework, mesosCloud.getMaster());
+          }
+          Status runStatus = driver.run();
+          if (runStatus != Status.DRIVER_STOPPED) {
+            LOGGER.severe("The Mesos driver was aborted! Status code: " + runStatus.getNumber());
+          }
+        } catch (RuntimeException e) {
```
vinodkone: why is a "try catch" changed sneaked in here? If it's not related to this change, please split that out into its own PR.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
Vlatombe: This is a common Jenkins plugin practice to retain backward compatibility, but at the same time migrate the old data to the new data structure. See https://wiki.jenkins-ci.org/display/JENKINS/Hint+on+retaining+backward+compatibility for details. I will add comments on the implementation details.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -114,42 +116,49 @@ public synchronized void init() {
...
-        + "\n" + "Checkpointing: " + framework.getCheckpoint()
-        );
-
-        if (StringUtils.isNotBlank(mesosCloud.getSecret())) {
+        try {
+          String targetUser = mesosCloud.getSlavesUser();
+          String webUrl = Jenkins.getInstance().getRootUrl();
+          if (webUrl == null) webUrl = System.getenv("JENKINS_URL");
+          StandardUsernamePasswordCredentials credentials = mesosCloud.getCredentials();
+          String principal = credentials == null ? "" : credentials.getUsername();
```
vinodkone: Previously, the default principal was "jenkins". Now it is being changed to empty string. This is not a backwards compatible change!

vinodkone:
There are some compatibility concerns with this change! Lets fix those.

More importantly, I don't understand the motivation for this change. Can you please include that in the description?

We definitely need feedback from the community regarding this change. Please send an email to the mailing list with the motivation and link to the PR.

Thanks.

src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
Vlatombe: I retain the fields as a matter of backward compatibility, otherwise it is much more complicated code-wise to migrate the existing data.

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
Vlatombe: https://github.com/jenkinsci/credentials-plugin/blob/master/src/main/java/com/cloudbees/plugins/credentials/CredentialsMatchers.java

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
Vlatombe: If any exception is raised during the thread execution, the thread just dies without any log. And worse, as the lines
```
driver = null;
running = false;
```
were not in a finally block, the scheduler would still be considered as running. Just set your authentication wrong once and you end up stuck.

Vlatombe:
Hi,

I think I have addressed your review comments, and I offloaded some items to other pull requests as requested.

brism:
:+1: 

If I understand this change correctly, it would solve a related issue I have with using the plugin in our environment. 

We're standing up a Jenkins master for any dev team that wants one, and they'll run slave jobs on Mesos. I would like to be able to give these teams Admin to configure the masters as they see fit.

The problem is that the **secret** for the Mesos cluster is right there *in plain text* on the Jenkins Config page. From a security perspective it would be much better if this were handled like a proper credential. I currently can't let them see this page or they'd have the creds for the Mesos cluster.

Thanks for the good work!

PR #182, opened 2015-11-04
Wrap JenkinsScheduler run body in a try/catch/finally.

Vlatombe:
Split from #179 

If any RuntimeException is raised during Scheduler execution, the object
was left in a corrupt state with running == true and driver != null.

The previous implementation would let any RuntimeException kill the
thread without any trace of it. Now, it is properly logged when it
happens and leave the scheduler in a correct state.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -114,42 +114,46 @@ public class JenkinsScheduler implements Scheduler {
     new Thread(new Runnable() {
       @Override
       public void run() {
-        String targetUser = mesosCloud.getSlavesUser();
-        String webUrl = Jenkins.getInstance().getRootUrl();
-        if (webUrl == null) webUrl = System.getenv("JENKINS_URL");
-        // Have Mesos fill in the current user.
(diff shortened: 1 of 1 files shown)


carlossg:
:bee:

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -114,42 +114,46 @@ public synchronized void init() {
...
+                driver = new MesosSchedulerDriver(JenkinsScheduler.this, framework, mesosCloud.getMaster());
+            }
+            Status runStatus = driver.run();
+            if (runStatus != Status.DRIVER_STOPPED) {
+                LOGGER.severe("The Mesos driver was aborted! Status code: " + runStatus.getNumber());
+            }
+        } catch(RuntimeException e) {
+            LOGGER.log(Level.SEVERE, "Caught a RuntimeException", e);
+        } finally {
+            driver = null;
```
vinodkone: i would also do a `driver.abort()` if `driver != null`.` this ensures this plugin doesn't hoard offers.

stephenc:
:bee:

Vlatombe:
@reviewbybees done

PR #185, opened 2015-11-18
update dependencies

flyinprogrammer:
not sure if helpful ¯\_(ツ)_/¯


flyinprogrammer:
huh those tests pass locally -- so that's weird. i'll look into it i guess.

Vlatombe:
Let's bump dependencies if we really have an added value for them.

PR #186, opened 2015-11-20
Move to Mesos 0.25.0

dberzano:


geoffroyjabouley:
As it been tested against a 0.25 Mesos cluster?
Maybe the official plugin should depend on a less recent version of Mesos, like 0.23.1.
Many features where introduce in the latest versions of Mesos, not sure the plugin is already using it.
Anyway, it is always possible to locally bump Mesos version and rebuild the plugin.

dberzano:
Hello @geoffroyjabouley, at CERN in the ALICE Experiment we are currently using version 0.24 (not 0.25) in production since its release. I'd feel confident about moving to 0.24 at least, I can't really say for 0.25.

geoffroyjabouley:
Hello
thanks for the feedback @dberzano.
I guess you have recompile the plugin file changing the pom.xml dependencie of Mesos to 0.24.
Good to know it is working. I will give it a try with different Mesos versions.

turb:
Note that there is a (tiny) breaking change in 0.26.
In ```JenkinsScheduler.java:146```, you have to remove ```ByteString.copyFromUtf8```.

Vlatombe:
Closing this pull request as it isn't necessary to rebuild the plugin as long as the signatures don't change.

PR #188, opened 2015-12-31
Modify JenkinsScheduler#findPortsToUse to return a SortedSet structure instead of a List.

Vlatombe:
This ensures port order is deterministic and should prevent any flaky behaviour for testing.

e.g. https://jenkins.ci.cloudbees.com/job/plugins/job/mesos-plugin/org.jenkins-ci.plugins$mesos/324/testReport/junit/org.jenkinsci.plugins.mesos/JenkinsSchedulerTest/testFindPortsToUseSamePortNumber/

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -31,6 +31,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -583,7 +579,7 @@ private void getContainerInfoBuilder(Offer offer, Request request, String slaveN
                           .setContainerPort(portMapping.getContainerPort()) //
                           .setProtocol(portMapping.getProtocol());
 
-                  int portToUse = portMapping.getHostPort() == null ? portsToUse.get(portToUseIndex++) : portMapping.getHostPort();
+                  int portToUse = portMapping.getHostPort() == null ? iterator.next() : portMapping.getHostPort();
```
oleg-nenashev: May throw runtime exception. Not a bug, because the original code behaves in such way (BTW Exception classes are different)

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -421,20 +423,15 @@ private boolean slaveAttributesMatch(Offer offer, JSONObject slaveAttributes) {
...
-          // If the port is already in the list of ports to use, ignore it and check the next one
-          while (candidatePort <= currentPortRange.getEnd() && portsToUse.size() < maxCount) {
-              if (!portsToUse.contains(candidatePort)) {
-                  portsToUse.add(candidatePort);
-              } else {
-                  candidatePort++;
-              }
-          }
+        // Check each port until we reach the end of the current range
+        int begin = (int) currentPortRange.getBegin();
```
oleg-nenashev: Why do we need to explicitly cast long to int here?

PR #189, opened 2016-01-04
Fix roundtrip on configuration if the label is set to empty string

Vlatombe:
If the label is set to empty string, when coming back to the configuration, the field is set to 'mesos'.

This change will set the value to mesos only if it is a new SlaveInfo

@reviewbybees esp. @carlossg


carlossg:
:bee: LGTM

PR #191, opened 2016-01-14
Set _JAVA_OPTIONS to use the executor memory as max heap

Vlatombe:
Recent JVMs determine automatically the max heap by taking about 1/4
of the available memory.

http://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/parallel.html#default_heap_size

The problem is that it takes as reference the host memory through
RLIMIT_AS flag value, and not the value defined by the cgroup it belongs
to.

As a consequence, if -Xmx is not explicitely specified and set to
relevant value, build processes may get killed by OOM-killer.

This takes the executor memory and set it as maximum heap size for any
forked jvm process which is what the user wants in most cases.


Vlatombe:
@reviewbybees

carlossg:
:bee:

Vlatombe:
@reviewbybees done

maselvaraj:
Thanks for the PR! :+1:

PR #194, opened 2016-01-19
Update jenkins slave naming to fix issues with spaces

sequences-io:
When jenkins slaves have multiple labels, there are spaces between the labels.  When this is translated directly into the mesos jenkins slave name it causes problems:

- When running slave.jar, jenkins does not enclose the name in quotes, so if there are spaces they are treated as separate command line parameters and this causes the java command to fail
- The mesos jenkins slave name is translated directly into a path so spaces and the ellipsis at the end make for difficult paths to work with.

This change removes any spaces from the label string and uses "left" instead of "abbreviate" to truncate the jenkins slave name to 63 characters instead of abbreviating it with an ellipsis.


Vlatombe:
Thanks for the fix!

PR #195, opened 2016-01-25
Fix the copy constructor

Vlatombe:
As credentialsId is @DataBoundSetter, it isn't part of the constructor, so it must be copied separately.

@reviewbybees


stephenc:
:bee:

carlossg:
:bee

Vlatombe:
@reviewbybees done

PR #198, opened 2016-01-27
Add ability to customize docker image using labels

itanf:


Vlatombe:
If I understand properly, this allows to configure a job to run using an arbitrary docker image, by using the label defined in slave info as a prefix, and reuse the rest of the SlaveInfo definition.

I don't think this can be merged as is. In my opinion, there should be an opt-in in the slave info to allow this behaviour. Not every Jenkins admin would like anyone creating jobs on their instance to be able to run an arbitrary docker image.

dirkhain:
That's a valid point @Vlatombe. Would a checkbox in the Cloud > Mesos-Cloud > Advanced section be suitable and sufficient for this? Default should be disabled and the plugin would check for this property before starting a container.  Please let me know what you think.

dirkhain:
@Vlatombe Could you give us some pointers how you envision the opt-in?

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -371,6 +371,9 @@ public boolean canProvision(Label label) {
         if (slaveInfo.matchesLabel(label)) {
           return true;
         }
+        if (label.toString().startsWith(slaveInfo.getLabelString() + ":")) {
```
Vlatombe: This test should be inside slaveInfo#matchesLabel

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -504,6 +507,15 @@ private MesosSlaveInfo getSlaveInfo(List<MesosSlaveInfo> slaveInfos,
       if (slaveInfo.matchesLabel(label)) {
         return slaveInfo;
       }
+      if (label.toString().startsWith(slaveInfo.getLabelString() + ":")) {
+        try {
+          String dockerImage = label.toString().substring(slaveInfo.getLabelString().length() + 1);
+          LOGGER.info("Using custom docker image: " + dockerImage);
```
Vlatombe: Reduce verbosity here.

Vlatombe:
Sorry for the delay. The opt-in should be at slave info level in my opinion.

PR #202, opened 2016-02-02
Fix getSlaveAttributeForLabel when labelName is null

fdonze:
Return slave attributes of a slave with no label (instead of null) when the request is for a slave with no label.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
@@ -517,14 +517,8 @@ public class MesosCloud extends Cloud {
 
   public JSONObject getSlaveAttributeForLabel(String labelName) {
     for (MesosSlaveInfo slaveInfo : slaveInfos) {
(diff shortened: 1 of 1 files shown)


fdonze:
@reviewbybees

carlossg:
:+1: LGTM

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -516,11 +516,15 @@ private MesosSlaveInfo getSlaveInfo(List<MesosSlaveInfo> slaveInfos,
   */
 
   public JSONObject getSlaveAttributeForLabel(String labelName) {
-    if(labelName!=null) {
-      for (MesosSlaveInfo slaveInfo : slaveInfos) {
+    for (MesosSlaveInfo slaveInfo : slaveInfos) {
+      if(labelName != null) {
```
Vlatombe: :ant: I'd use StringUtils.equals for clarity.

PR #206, opened 2016-02-16
Add escape-by-default to jelly files

carlossg:
Test fail in newer jenkins versions without it

@reviewbybees


Vlatombe:
:bee:

maselvaraj:
Thanks for the PR!

jglick:
Might as well just upgrade to the 2.x parent POM. If you do that, `mvn test` will actually enforce this fix.

PR #207, opened 2016-02-16
equals method is wrong, does not consider object fields

carlossg:
Same thing goes for hashCode, althugh not wrong can be improved to consider all fields

@reviewbybees


stephenc:
:bee:

Vlatombe:
:bee:

carlossg:
@reviewbybees done

PR #208, opened 2016-02-17
Slaves should not try to reconnect

carlossg:
We have seen instances where the slave info is deleted from the master and the slave keeps running forever
probably due to some race condition
Default to no reconnect, but can be overriden

@reviewbybees


stephenc:
:bee:

andresrc:
:bee:

carlossg:
@reviewbybees done

PR #210, opened 2016-02-26
Update mesos dependency to 0.27.0

Vlatombe:
We require now mesos 0.27.0 to build the plugin but we retain runtime compatibility with versions older than 0.25.0 by using reflection for the changed API.

External containerizer has been deprecated in Mesos, so support has been removed from the plugin.

@reviewbybees


goettl79:
Would also appreciate an update to 0.27 and the removal of deprecated functionality.

PR #211, opened 2016-03-01
Wait for slave to come online to avoid provisioning of additional slaves

fdonze:
If the slave is not connected, Jenkins might try to launch additional slaves which results in over-provisioning slaves and reserving additional resources in the mesos cluster.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosComputerLauncher.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosComputerLauncher.java
@@ -119,7 +119,7 @@ public class MesosComputerLauncher extends ComputerLauncher {
       try {
         LOGGER.info("Waiting for slave computer connection " + name);
         Thread.sleep(5000);
-      } catch (InterruptedException ignored) {}
+      } catch (InterruptedException ignored) { return; }
     }
     if (computer.isOnline()) {
       LOGGER.info("Slave computer connected " + name);
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosComputerLauncher.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosComputerLauncher.java
@@ -107,22 +107,22 @@ public class MesosComputerLauncher extends ComputerLauncher {
       // Since we just launched a slave, remove it from pending deletion in case it was marked previously while
       // we were waiting for resources to be available
       computer.getNode().setPendingDelete(false);
-      waitForSlaveConnection(computer);
+      waitForSlaveConnection(computer, logger);
       logger.println("Successfully launched slave" + name);
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/mesos/MesosComputerLauncher.java
```diff
@@ -107,12 +107,27 @@ public void failed(JenkinsSlave slave) {
...
       logger.println("Successfully launched slave" + name);
     }
 
     LOGGER.info("Finished launching slave computer " + name);
   }
 
+  private void waitForSlaveConnection(MesosComputer computer) {
+    while (computer.isOffline() && computer.isConnecting()) {
+      try {
+        LOGGER.info("Waiting for slave computer connection " + name);
```
Vlatombe: I think 'FINE' would be enough otherwise it will be too verbose for regular use.

src/main/java/org/jenkinsci/plugins/mesos/MesosComputerLauncher.java
```diff
@@ -107,12 +107,27 @@ public void failed(JenkinsSlave slave) {
...
+        Thread.sleep(5000);
+      } catch (InterruptedException ignored) { return; }
+    }
+    if (computer.isOnline()) {
+      LOGGER.info("Slave computer connected " + name);
```
Vlatombe: Same as above.

Vlatombe:
:bee:

stephenc:
I would like to review this change on a big screen before you look to merge. It smells risky from looking at it on my phone

Vlatombe:
@stephenc if you want to put your hand in the honey jar and hint us towards the proper fix for this, you're more than welcome!

stephenc:
That is my intent but it will be tomorrow am before I can deliver on said intent ;-)

stephenc:
:bee: this is good, but likely not the whole story...

stephenc:
The full fix would be to reduce the excess workload by the number of nodes that are provisioned but not connected and will match the label... for fellow bees [here is an example](https://github.com/cloudbees/operations-center-cloud/blob/master/src/main/java/com/cloudbees/opscenter/client/cloud/CloudImpl.java#L101-L122) of doing just that where we count up the actual pending workload capacity and subtract that from the excess workload.

This is necessary because only a cloud can know if a slave is actually trying to connect or has given up and thus the lack of executors should be considered as excess workload.

Older Jenkins cores (pre 1.609) would suffer really badly without this finessing. Post 1.609 the finesse *should* be unnecessary, however depending on how the slaves connect in, a cloud may need to have its own special handling anyway to prevent against excess provisioning.

I'd run with these changes for now and if that still is triggering over-provisioning then replicate my tricks

Issue #204, opened 2016-02-03
Error in getWorkspace while running polling source code manager


jAlpedrinha:
I'm running a DCOS (v0.3.0) cluster with Jenkins installed, and I'm trying to trigger builds on git repository changes, but during polling every now and then I get an error on Polling Logs.

Stacktrace:
```
Started on Feb 3, 2016 3:48:35 PM
ERROR: Failed to record SCM polling
java.lang.NullPointerException
	at org.jenkinsci.plugins.mesos.MesosWorkspaceBrowser.getWorkspace(MesosWorkspaceBrowser.java:38)
	at hudson.model.AbstractProject._poll(AbstractProject.java:1395)
	at hudson.model.AbstractProject.poll(AbstractProject.java:1355)
	at com.cloudbees.jenkins.GitHubPushTrigger$1.runPolling(GitHubPushTrigger.java:84)
	at com.cloudbees.jenkins.GitHubPushTrigger$1.run(GitHubPushTrigger.java:110)
	at hudson.util.SequentialExecutionQueue$QueueEntry.run(SequentialExecutionQueue.java:118)
	at jenkins.util.ContextResettingExecutorService$1.run(ContextResettingExecutorService.java:28)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

I'd be more than glad to share any more details you require.

maselvaraj:
Can you please post your Jenkins - mesos cloud config ?

jAlpedrinha:
Sure! 
I believe I understood the issue now, when polling tasks are performed, there are no executors available, since there is no slave working, and the master has 0 executors so the mesos plugin can't provide a workspace. I found a workaround for this issue by enabling one executor on the master node, which will be able to perform the polling tasks, instead of relying on slaves. Additionaly I set a tag on slaves so that build jobs can only run on slaves (avoiding using master to perform such tasks).
I still believe this is a bug, because when there are no executors available polling doesn't work, and the message could be slightly better. This could be solved by either allowing a special worker to perform such tasks, or by triggering a dedicated slave for this. 

Any thoughts? 

[...]

PR #212, opened 2016-03-02
introduced affinity in offer selection

sathiya-mit:


sathiya-mit:
@maselvaraj @vinodkone Please review it and let me know your valuable comments.

sathiya-mit:
Along with this pull request, I have provided a fix for https://github.com/jenkinsci/mesos-plugin/issues/204

Vlatombe:
Do you mind providing your NPE fix in a separate pull request? This has nothing to do with the current PR.

sathiya-mit:
sure will do...

maselvaraj:
@Vlatombe this pull request bumps up the most recently used mesos slave in the offer list[if it is available]. If the job goes to the same slave once again, build time will be faster than going to a new slave. Thoughts ?

Vlatombe:
I think the intent is ok, but why limit ourselves to the latest slave? I'd store a LRU of slave id instead rather than the hostname.

sathiya-mit:
@Vlatombe Thanks for your comment. Just to make it simple. I did this way.
Do you want to store all the used slave ids in a LRU fashion, and juggle the entire offer list to most recently used order ?

Vlatombe:
I would store the 10 latest slaves id (size configurable through a system property) and prioritize these.

sathiya-mit:
@Vlatombe Thanks for your quick response. I will make the necessary change.

sathiya-mit:
Implemented LRU as you suggested. Please let me know your thoughts.

geoffroyjabouley:
Hello

can you please provide more details on how this is intended to speed up
builds?

Are you mounting on each jenkins slave a workspace volume local on each
mesos slave machine?

2016-03-05 16:34 GMT+01:00 Sathiya Narayanan <notifications@github.com>:

sathiya-mit:
@geoffroyjabouley Take the case of maven job, if we provide a recent slave (where the job has already ran) we can prevent downloading the maven dependency jars/artifacts again. This will help us in improving the maven build time.

Nope, We don't mount workspace volume on each mesos slave machine.

geoffroyjabouley:
Thanks
your use case seems to indicate that you execute your jobs in jenkins
slaves running directly on the mesos slave machine, without docker
containerizer. Am i right?

2016-03-06 18:45 GMT+01:00 Sathiya Narayanan <notifications@github.com>:

sathiya-mit:
@geoffroyjabouley Yes, You are right.

maselvaraj:
@geoffroyjabouley It makes sense only if Mesos plugin is configured with custom remoteFS root. That is the Jenkins slave provisioned  through Mesos plugin will use the same path for workspace in any mesos slave.

sathiya-mit:
@Vlatombe thanks for reviewing and merging it :+1:

PR #215, opened 2016-03-04
Fix another case of duplicate framework registration

Vlatombe:
Added a Garbage Collector to stop the scheduler and remove the mesos framework in case the cloud gets removed.

Doing so, I found out that there was a problem related to credentialsId and usage of @DataBoundSetter : when the scheduler is restarted after constructor call (onDemandRegistration=false), credentialsId value is not yet set and this can cause a duplicate framework registration. So the field initialization has been moved to constructor.

Last, added equals/hashCode implementations to MesosSlaveInfo and embedded classes, as these are necessary to compare MesosCloud instances, and these are used in Mesos#getInstance to prevent building duplicate MesosImpl instances.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/Mesos.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/Mesos.java
@@ -14,10 +14,16 @@
  */
 package org.jenkinsci.plugins.mesos;
 
+import hudson.Extension;
+import hudson.XmlFile;
+import hudson.model.Saveable;
+import hudson.model.listeners.SaveableListener;
+import jenkins.model.Jenkins;
 import org.apache.mesos.Scheduler;
 
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 
 public abstract class Mesos {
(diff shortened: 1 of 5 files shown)


fdonze:
:bee:

src/main/java/org/jenkinsci/plugins/mesos/Mesos.java
```diff
@@ -68,7 +74,13 @@ public SlaveRequest(JenkinsSlave slave, double cpus, int mem, String role,
...
   abstract public void updateScheduler(String jenkinsMaster, MesosCloud mesosCloud);
   abstract public boolean isSchedulerRunning();
-  abstract public void stopScheduler();
+
+  /**
+   * Stops the scheduler.
+   * @param force {@code false} to stop the scheduler gracefully, {@code true} to force stop
+   * @return {@code true} if the scheduler is stopped after the call, {@code false} otherwise.
+     */
+  abstract public boolean stopScheduler(boolean force);
```
carlossg: any concern about breaking the API? could easily keep the old method

src/main/java/org/jenkinsci/plugins/mesos/Mesos.java
```diff
@@ -104,4 +116,25 @@ public static synchronized Mesos getInstance(MesosCloud key) {
...
+   * When Jenkins configuration is saved, teardown any active scheduler whose cloud has been removed.
+   */
+  @Extension
+  public static class GarbageCollectorImpl extends SaveableListener {
+
+    @Override
+    public void onChange(Saveable o, XmlFile file) {
+      if (o instanceof Jenkins) {
+        Jenkins j = (Jenkins) o;
+        for (Iterator<Map.Entry<MesosCloud, Mesos>> it = clouds.entrySet().iterator(); it.hasNext();) {
```
carlossg: if you have another cloud wouldn't this throw a casting exception?

Vlatombe: where do you see a cast?

src/main/java/org/jenkinsci/plugins/mesos/Mesos.java
Vlatombe: true, i'll do that

src/main/java/org/jenkinsci/plugins/mesos/Mesos.java
carlossg: never mind, I thought clouds had all the clouds, not just Mesos

PR #217, opened 2016-03-10
Add an administrative monitor to notify if mesos slave provisioning couldn't be done.

Vlatombe:
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -354,10 +354,9 @@ public class JenkinsScheduler implements Scheduler {
   }
 
   /**
-   * Returns the unmatched labels during the last call of resourceOffers
-   * @return the unmatched labels during the last call of resourceOffers
+   * @return the unmatched labels during the last call of resourceOffers.
      */
-  public Set<String> getUnmatchedLabels() {
+  public synchronized Set<String> getUnmatchedLabels() {
     return Collections.unmodifiableSet(unmatchedLabels);
   }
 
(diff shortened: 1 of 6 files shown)


Vlatombe:
Here is a sample as screenshot

capture d ecran 2016-03-10 a 18 00 49

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -343,6 +348,17 @@ public synchronized void resourceOffers(SchedulerDriver driver, List<Offer> offe
...
+    for (Request request: requests) {
+      unmatchedLabels.add(request.request.slaveInfo.getLabelString());
+    }
+  }
+
+  /**
+   * Returns the unmatched labels during the last call of resourceOffers
+   * @return the unmatched labels during the last call of resourceOffers
+     */
+  public Set<String> getUnmatchedLabels() {
```
oleg-nenashev: Missing synchronization?

src/main/resources/org/jenkinsci/plugins/mesos/MesosAdministrativeMonitor/message.jelly
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:i="jelly:fmt" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+    <div class="warning">
+        <p>Some Mesos slaves cannot be provisioned on the following labels:</p>
+        <ul>
+            <j:forEach items="${it.labels}" var="label">
+                <li><b>${label}</b></li>
```
oleg-nenashev: what about hyperlinks to label expressions? It allows to get usage stats for free

carlossg:
LGTM :bee:

PR #218, opened 2016-03-11
Add inline help and removed unnecessary field entries.

Vlatombe:
@reviewbybees


armfergom:
:bee: Found no `help-{field}.html` for fields `role`and `slavesUser` (mentioning in case they were forgotten)

oleg-nenashev:
Looks good to me. BTW seems this check implies bugfixes for null instance

PR #229, opened 2016-04-06
jenkins slave memory fix

zazrivec:
Jenkins-slave  can consume all of allocated memory via -Xmx switch. Bugfix:. Only slave memory amount should be assigned.


Vlatombe:
Good catch! Thank you

Issue #232, opened 2016-04-14
Enhancement - Create a Jenkins slave with at least N executors


geoffroyjabouley:
There is currently an option to configure the Maximum number of Executors per Slave.

What about another option to configure the Minimum number of Executors per Slave?

My use case is the following:
i'm starting a job, running on a jenkins mesos slave COMMON, which in turn starts a new job, then waits for its completion. Both jobs are "generic" and could be executed by the same COMMON slave if it was possible , saving time and resources. But as they are not enqueued at the same time, Mesos cloud plugin will never create a COMMON slave with 2 executors.

So i would like to configure that mesos jenkins slave COMMON is always created with at least 2 executors (and possibly up to 4).

Do you think it is possible?

geoffroyjabouley:
For the record, such "over-provisioning" shall not affect the remaining Workload, and the minExecutors variable only be used when provisioning the slave https://github.com/jenkinsci/mesos-plugin/blob/master/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java#L415

doProvision(Math.max(numExecutors, slaveInfo.getMaxExecutors()), slaveInfo);


Not sure about impact on the Single-Use slave feature.
ps: BTW, how the provisioning of a Single-Use slave with 2 executors affect the remaining workload of Jenkins?

PR #233, opened 2016-04-20
Add a minimum number of executors per slave #232

etiennebec:
Current behavior won't change (default value set to 1) and the update of the workload has been modified to take in account this parameter.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
@@ -44,6 +44,7 @@ public class MesosSlaveInfo extends AbstractDescribableImpl<MesosSlaveInfo> {
   private final double slaveCpus;
   private final int slaveMem; // MB.
   private final double executorCpus;
+  private final int minExecutors;
   private final int maxExecutors;
   private final int executorMem; // MB.
   private final String remoteFSRoot;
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -152,6 +156,10 @@ public MesosSlaveInfo(
     this.additionalURIs = additionalURIs;
     this.nodeProperties.replaceBy(nodeProperties == null ? new ArrayList<NodeProperty<?>>() : nodeProperties);
 
+    // Ensure minExecutors is at least equal to 1 and prevent case where minExecutors > maxExecutors.
+    int minExecutorsVal = Integer.parseInt(minExecutors);
+    this.minExecutors = minExecutorsVal < 1 || minExecutorsVal > this.maxExecutors ? 1 : minExecutorsVal;
```
Vlatombe: if minExecutorsVal > this.maxExecutors => should be set to maxExecutors, not 1.

etiennebec: I think min > max is an anormal case, most likely the result of a typing error when configuring a slave. In my opinion the safest fallback is to set the value to 1 so the admin might notice the wrong number of executors but without provisioning the unnecessary resources that the maxExecutors value would lead to.

vinodkone:
If min > max is unexpected, lets just do a validation at the input time?

etiennebec:
I will add a doCheck to prevent wrong values to be entered, but I will keep the minExecutorsVal < 1 check in MesosSlaveInfo so when updating the plugin on an existing Jenkins, the parameter won't be set at 0 but at 1 (after one save/reboot).

etiennebec:
I added the doCheck methods on min and max executors fields.

PR #240, opened 2016-06-20
Added a check to see if there is a build in queue before declining an…

maselvaraj:
… offer

Don't decline the offer is there is a build in Jenkins queue. There may be a scenario where the slave provisioning has not started yet, but there is a build in queue.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -18,13 +18,14 @@ package org.jenkinsci.plugins.mesos;
 
 import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
 import com.google.common.annotations.VisibleForTesting;
+
 import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.model.Item;
 import hudson.model.Computer;
 import hudson.model.Node;
 import hudson.util.Secret;
 
 import java.lang.Math;
-
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -483,6 +484,19 @@ private boolean matches(Offer offer, Request request) {
     }
   }
 
+  private boolean buildsInQueue(hudson.model.Queue queue) {
+   hudson.model.Queue.Item[] items =  queue.getItems();
```
Vlatombe: I'd prefer something like getApproximateItemsQuickly() to prevent thread contention.

maselvaraj: @Vlatombe looks like getApproximateItemsQuickly is deprecated...

maselvaraj: getApproximateItemsQuickly()
Deprecated. Use getItems() directly. As of 1.607 the approximation is no longer needed.
Like getItems(), but returns an approximation that might not be completely up-to-date.

vinodkone:
@maselvaraj I'm curious why you made this change. Have you seen any issues in your cluster without this fix? The think just looking `requests` structure should be good enough; looking at build queue is misleading because that queued build might just land on an existing jenkins slave. I've recently implemented suppression and I see that Jenkins is not suppressing because there are builds in the queue, even though the build is queued up on an existing executor. I plan to revert this change. Please let me know if you have any objections.

maselvaraj:
@vinodkone While Jenkins slave provisioning algorithm(by default) makes best use of a connected/online slave before deciding to provision the next slave,  in our setup, we want builds to be picked up immediately (queue time as minimum as possible) so we have fine tuned node  provisioner margin accordingly (https://support.cloudbees.com/hc/en-us/articles/115000060512-New-Shared-Agents-Clouds-are-not-being-provisioned-for-my-jobs-in-the-queue-when-I-have-agents-that-are-suspended).  Which is why we don't want offers to stop until builds are in queue.

But I see your point, wondering if we can keep this option as a checkbox/optional ?

vinodkone:
@maselvaraj Thanks for the quick reply and context. Really helpful. 

IIUC, you updated the margin parameters for node provisioner so that new agents are provisioned quickly right? That means those agent provisioning requests should be queued in `requests` queue of the scheduler pretty quickly. So why is that not enough?

maselvaraj:
@vinodkone Yes, that is right. As long as the reviveOffers() is delivered to mesos on a new request, we won't need this. You can remove this.

vinodkone:
Great. Thanks.

vinodkone:
Fix https://github.com/jenkinsci/mesos-plugin/pull/354

Issue #216, opened 2016-03-07
Support for Maintenance Primitives


nnordrum:
I wanted to get a conversation started on supporting [Maintenance Primitives](http://mesos.apache.org/documentation/latest/maintenance/) in the Framework. They are discussed in this [Mesos blog post](https://mesosphere.com/blog/2015/10/07/mesos-inverse-offers/).

> ## Maintenance-ready frameworks
> 
> One important note before we continue: the interface for inverse offers was not added to the native C++ scheduler interface. Instead, a framework must use the V1 HTTP API to receive inverse offers.
> 
> For maintenance primitives, inverse offers play a key role in draining a machine of tasks prior to starting maintenance on a machine. Draining a machine means stopping all the tasks running on the machine prior to the maintenance event. Frameworks are responsible for draining machines when instructed by the Mesos master (if they don’t, the operator can still “hard drain” the machine by pulling the plug).
> 
> In the workflow provided by the maintenance primitives feature, a machine starts to drain resources as soon as it is scheduled for maintenance. (For more details on how schedules are constructed, see the maintenance documentation). All frameworks using resources on the draining machine receive inverse offers periodically, asking for the resources back.

I have yet to find an example of how to have your framework register/talk to/etc. master using the .so _and_ the HTTP API, so that's probably step one.

Just starting the conversation, so, GO!

vinodkone:
Mesos maintenance primitives have not been ported to the SchedulerDriver interface, but you can get access to them with new HTTP API interface. So, step 1) is likely to port Jenkins Scheduler to use the new HTTP API interface.

nnordrum:
I got the impression reading the docs that it's not an either/or situation, but not _sure_ one way or the other. I think that becomes step 1 :)

nnordrum:
[Here's a reactive library that may be a good plugin](https://github.com/mesosphere/mesos-rxjava)

jeschkies:
[USI](https://github.com/mesosphere/usi) uses the V1 API. Please ask for maintenance support there.

PR #242, opened 2016-06-23
Simple "Maintenance Mode" Implementation (not using Maintenance Primitives)

nnordrum:
#216

This is a "better than nothing" change until the Maintenance Primitives can be used, but that's going to be a while.

Added a decline hosts setting to pass in a CSV of regex patterns to match hosts that you don't want jobs scheduled on. Primary use case is maintenance mode. 

Default value, blank, has no change in behavior.

Also made a change in MesosSlaveInfo that assumes you're using a docker container for your slave.

```
if (!containerInfo.getDockerImageCustomizable()) {
```

became
    if (containerInfo == null || !containerInfo.getDockerImageCustomizable()) {

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -22,37 +22,12 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.model.Computer;
 import hudson.model.Node;
 import hudson.util.Secret;
-
-import java.lang.Math;
-
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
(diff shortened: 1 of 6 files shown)


janisz:
@nnordrum nice idea but in this approach you ask user to duplicate maintenance definitions in Mesos and Jenkins. How about reading maintenance schedule from Mesos itself? For example periodically check [`/maintenance/schedule`](https://mesos.apache.org/documentation/latest/maintenance/) and update Jenkins excluded hosts. https://github.com/mesosphere/marathon/issues/2414#issuecomment-254196961

brugidou:
Hi @nnordrum I agree with @janisz using maintenance schedule from Mesos would be really good!

nnordrum:
I agree with you, @brugidou, and @janisz. My understanding at the time was that the API we were using at the time did not support the maintenance primitives, hence the workaround. I'm going to close this.

PR #243, opened 2016-06-24
Introduce support for Mesos maintenance primitives

ederst:
This commit introduces the feature for the frameworks registered by Jenkins to ignore
offers having an unavailability (maintenance primitive) set.

The offers are only ignored when the current time is between the start and end time (start + duration)
of the unavailability.

Issue: https://github.com/jenkinsci/mesos-plugin/issues/216

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -29,6 +29,7 @@ import java.lang.Math;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


nnordrum:
did I miss an update from Mesos where they _are_ planning on supporting Maintenance Primitives in the .so? Last I checked, which was a while ago, they were only going to support Maintenance Primitives in the HTTP API.

ederst:
IDK. I read the posts in https://github.com/jenkinsci/mesos-plugin/issues/216 too and saw that it is not supported. However, I debugged the code and when I schedule a maintenance on my 0.27.2 Mesos  cluster the offers for the specified host(s) in the schedule contain an `Unavailability` object when entering `resourceOffers()`.

nnordrum:
Interesting--well it that's the case, I would likely prefer this over #242.

ederst:
One thing I found out during testing: It is not sufficient to just provide "host". The primitives only work if I specify the "ip" of the slave, too.

ederst:
So I've tried this with the code of the pull request as well, and it works just fine. What this means is, that the plugin will not schedule a task/Jenkins slave on a Mesos agent being scheduled for downtime.

However, I forgot that this version of the plugin keeps Jenkins slaves online for a given time (deletes them when reaching an idle timeout and in the mean time utilizes them). Which means that potentially this could lead to an awfully long maintenance, waiting for the Jenkins slaves to vanish and the Mesos agents to drain.

Therefore, to improve and prevent this, the code would be better with some kind of mechanism to suspend/remove the Jenkins slaves of a Mesos agent which is currently in "draining_mode".

nnordrum:
Well, "drain mode" is optional, as Frameworks aren't _required_ to fully drain for a maintenance period.

I agree it would be nice if it also drained idle/unschedulable slaves, but as long as it doesn't schedule new things on them, I'd be fine with that, at least for now.

I might also add another parameter of "don't schedule on node _N_ minutes before maintenance is scheduled to start" with a default of like 15.

ederst:
> I might also add another parameter of "don't schedule on node N minutes before maintenance is scheduled to start" with a default of like 15.

Yeah, I don't know if this is necessary. You could simply schedule the maintenance window accordingly, so I don't think it is worth the hassle.

Btw. this still works with Mesos 1.2.0.

maselvaraj:
Thanks @ederst for the PR! LGTM. Can you please add unit tests to cover this ?

maselvaraj:
Thanks @ederst for the PR and the unit tests !

PR #246, opened 2016-07-06
Fix: Modifying MesosCloud Configuration Kills Running Jobs

titosand:
Saving a new MesosCloud config causes all running tasks to terminate because the Jenkins scheduler is being needlessly restarted upon every config change. These changes address this problem and allow the Jenkins scheduler to stay running when config changes are made to the MesosCloud.

The root cause of this issue was the GarbageCollector in Mesos.Java incorrectly removing modified (not removed) clouds - this is a consequence of using mutable keys in a HashMap. As a fix, I have implemented a cloudID that serves as a unique identifier for each MesosCloud. This allows for "modified" clouds (i.e. same cloud with a changed description or slaveInfo) to be recognized and not accidentally removed by the GarbageCollector.

I also fixed a NPE occuring in the matchesLabel method.


Vlatombe:
Looks good to me. Thank you!

PR #275, opened 2016-12-09
Global configuration for making single use slave default for all the …

maselvaraj:
…jobs


AlanFoster:
@maselvaraj This seems like a useful configuration to have! Is this something that could be merged in perhaps? 🤔 

I'm wishing to have single instance slaves within Jenkin's declarative pipelines, and this would be a nice configuration value to set globally if possible :+1:

PR #277, opened 2016-12-21
Pipeline support for MesosSingleUseSlave

### LINKED ISSUE #262: Single use slave in piepeline
When setting up a 'standard' job in jenkins, one can tick the 'Mesos Single-Use Slave' box so that only one job can get executed on a slave before the slave gets deleted.

When setting up a job with the new 'pipeline' feature, one has to write the job instructions in a 'node' block :
    node('node_label) {
    }

Is is possible to specify the 'Mesos Single-Use Slave' thing in a pipeline job ?
I switched to pipeline jobs recently and the nice isolation I had before is gone now.

Thanks

varju:
Convert MesosSingleUseSlave to a SimpleBuildWrapper so that it may be used from Pipeline scripts

Fixes #262


maselvaraj:
Thank you @varju for the PR!

PR #285, opened 2017-03-02
Ensure task is launched with appropriate roles on ports from offer.

damomurf:
If a Mesos scheduler is configured with a role, the scheduler receives offers for resources under that defined role (or the default role).

Currently, the code that builds tasks for execution doesn't include any role attached in the offer for assigned port resources.

This PR ensures that the role for port resources on the offer is used, just like it is for cpu and mem.

This problem was manifesting itself with errors in both the Jenkins and mesos-master logs and a number of failed Jenkins slaves would appear in the Jenkins UI, e.g.:

```
Sending status update TASK_ERROR for task mesos-jenkins-506c01240453427bb20f1d3dff26b718-mesos-ultra-e2e of framework 7ef79e0d-1cc1-4880-ad98-7fb0329fa816-0003 'Total resources cpus(jenkins_sg):1.1; mem(jenkins_sg):6899.2; ports(*):[31000-31000] required by task and its executor is more than available cpus(jenkins_sg):29.4; mem(jenkins_sg):118622; disk(jenkins_sg):297124; ports(jenkins_sg):[31000-32000]'
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -98,6 +98,7 @@ public class JenkinsScheduler implements Scheduler {
       "java -DHUDSON_HOME=jenkins -server -Xmx%dm %s -jar ${MESOS_SANDBOX-.}/slave.jar %s %s -jnlpUrl %s";
   private static final String JNLP_SECRET_FORMAT = "-secret %s";
   public static final String PORT_RESOURCE_NAME = "ports";
+  public static final String MESOS_DEFAULT_ROLE = "*";
 
   private Queue<Request> requests;
   private Set<String> unmatchedLabels;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -753,6 +768,7 @@ private void getContainerInfoBuilder(Offer offer, Request request, String slaveN
                   .newBuilder()
                   .setName("ports")
                   .setType(Value.Type.RANGES)
+                  .setRole(roleToUse)
```
maselvaraj: What would happen if there no role set at Mesos ? In case if it returns null will this affect this ?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -551,6 +551,20 @@ void setDriver(SchedulerDriver driver) {
   }
 
   @VisibleForTesting
+  String findRoleForPorts(Offer offer) {
+
+      String role = null;
```
maselvaraj: This might return null. I was wondering if setting 'null' for role might affect the task launch

damomurf: We could set the initial value to "*" which is what the old functionality was anyway? That would mitigate the concern with the use of the return value where this is used on line 771?

maselvaraj: Yes, Default value '*' sounds okay to me.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
damomurf: I've amended the PR to use the default value for the mesos role, in case nothing is set by Mesos.

maselvaraj:
Thanks @damomurf for the PR!

PR #293, opened 2017-05-09
add hostname to offer attributes

fugi:
This add _hostname_ to offer attributes. Also adds possibility to use regex matching.


maselvaraj:
@fugi Currently if you set an attribute at Mesos for hostname it can be used in Jenkins mesos plugin. I was wondering if this will be confusing if there is an attribute called 'hostname' at Mesos slave.

vinodkone:
Closing since this has gotten stale. Please re-open if you are still interested in getting this merged.

PR #294, opened 2017-06-14
Add option to specify disk resource.

rmathure:
This option could be customized at slave level and hence
plugin would only accept resources that has the specified
amount in MB as Disk resource available. The option is
ignored for Docker containerizer and if the value is set
to 0.

This opens up up the killing of mesos slave if the slave
uses more that alloted amount of disk space by setting
--isolation with disk/du and
--enforce-container-disk-quota


maselvaraj:
@rmathure Thanks for the PR!

PR #299, opened 2017-07-24
simplify slaveAttributesMatch

chrisliu529:
collect offer attributes only when slaveAttributes exist


jeschkies:
@chrisliu529, thanks for you request. Is it still required?

jeschkies:
@chrisliu529, I just stumbled upon the matching again. Would you mind fixing the diff?

PR #300, opened 2017-08-11
Allow tunnel option in slave config

LewG02:
Currently, if you use the plugin you're limited to going to the default setup for a JSlave.  That doesn't work in all scenarios, especially ours.  We need to add in the Jenkins tunnel parameter to the JSlave that is started.  See JENKINS_TUNNEL for this Docker image for more infromation: https://hub.docker.com/r/jenkinsci/jnlp-slave/


jeschkies:
@ghost, this request is very old. Do you think this is still required?

Issue #251, opened 2016-07-27
Docker slaves launched with invalid hostname when label contains underscores


thecubed:
Hi all,

It would appear that if the label string for a Docker slave contains an underscore (i.e. `j_node5`) , Mesos tries to launch the Docker image with `--hostname mesos-jenkins-(UUID)-j_node5`.

Docker then complains that this is an invalid hostname and fails to launch the Jenkins node.
(This continues forever, filling up Jenkins with slaves that don't exist in Mesos)

A simple fix would be to strip the hostname of characters that aren't allowed as hostnames here: https://github.com/jenkinsci/mesos-plugin/blob/70fe128ca0d0e3f5d5718dbd9e8953133096331e/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java#L708 (or so it appears).

Thanks!

matt2000:
I'm having the same symptoms, but removing the underscore from my label did not resolve the issue. How do you know that "Docker the complains that this is an invalid hostname" ? Where can I find the logs indicating this? I see no such thing in the jenkins node launch log, before or after changing the label.

thecubed:
The error was visible in the Mesos webui for the task. If you look at the bottom half of your Mesos webui, you should see the failed/finished tasks, and the Jenkins launch attempts should be there. The error was visible in one of the stderr/stdout links in the sandbox page.

It's also possible that your node name makes the Docker hostname too long - I think Docker enforces the 64-character limit on hostnames, so if the strings "mesos-jenkins-(whatever uuid here)-(your slave label here)" is too long it'll fail to launch as well.

epkabeg:
We are having issues with this as well. Since a recent version of Docker, it enforces proper hostnames.

When the plugins is configured with "Docker Image Can Be Customized" it adds a : to the hostname so this functionality is broken with newer Docker versions.

sbeaulie:
Did anyone find a workaround for this? Our mesos slave label ends up longer than 64, contains hyphens but no underscore and contains ":" to specify the image version we want

jeschkies:
I think this has been fixed in #367.

Issue #295, opened 2017-06-29
Can't customize container with private image


michaeljs1990:
When I select the "Docker Image Can Be Customized" checkbox the container constantly fails even when the container I am overwriting it with is the default image that is working for other jobs.

This is all that shows up on the slave when it fails
```
Jun 29 12:43:08 ms-1 mesos-slave[11508]: E0629 16:43:08.662935 11565 slave.cpp:3903] Failed to update resources for container c683e7a4-db0f-4ba2-a5ad-dbc57fd5c3d4 of executor 'mesos-jenkins-2ed8308e20af4dcfb63d15b300797c1a-mesosBase:docker' running task mesos-jenkins-2ed8308e20af4dcfb63d15b300797c1a-mesosBase:docker on status update for terminal task, destroying container: Failed to run 'docker -H unix:///var/run/docker.sock inspect mesos-85f895ed-06ee-401c-96c9-42dc051afd76-S1.c683e7a4-db0f-4ba2-a5ad-dbc57fd5c3d4': exited with status 1; stderr='Error: No such image or container: mesos-85f895ed-06ee-401c-96c9-42dc051afd76-S1.c683e7a4-db0f-4ba2-a5ad-dbc57fd5c3d4
Jun 29 12:48:02 ms-1 mesos-slave[11508]: I0629 16:48:02.811250 11591 slave.cpp:4952] Framework 85f895ed-06ee-401c-96c9-42dc051afd76-1473 seems to have exited. Ignoring registration timeout for executor 'mesos-jenkins-2ed8308e20af4dcfb63d15b300797c1a-mesosBase:docker'
Jun 29 12:48:29 ms-1 mesos-slave[11508]: I0629 16:48:29.714737 11575 slave.cpp:5044] Current disk usage 9.71%. Max allowed age: 5.620507282549398days
Jun 29 12:49:00 ms-1 mesos-slave[11508]: I0629 16:49:00.600776 11577 slave.cpp:4952] Framework 85f895ed-06ee-401c-96c9-42dc051afd76-1473 seems to have exited. Ignoring registration timeout for executor 'mesos-jenkins-dad3577052e549f2810441706719b740-mesosBase:docker'
```

My jenkinsfile looks mostly like this...

```
node ('mesos:docker.int.company.net/base/image:latest') {...}
```

Where `docker.int.company.net/base/image:latest` is a known working image that I am currently running other jobs on.

```
[...]

PR #302, opened 2017-09-06
Replace invalid hostname characters for task names and docker hostname

sbeaulie:
Before this change, if you used the mesos plugin option 'Docker Image Can Be Customized'
and the resulting task name was less than 64 characters it could happen that the docker
image name contained invalid characters both for the task name and the docker hostname.
For example a custom label:foo/bar:latest would fail as a task name because / is invalid
in a task name since it breaks the directory structure. It would also fail as a docker
hostname since it needs to comply with RFC 1123 hostnames

Original commit was to fix https://github.com/jenkinsci/mesos-plugin/issues/251

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -243,10 +243,26 @@ public class JenkinsScheduler implements Scheduler {
         return prefix + '/' + suffix;
     }
 
-    public synchronized void terminateJenkinsSlave(String name) {
-        LOGGER.info("Terminating jenkins slave " + name);
+  /**
+   * When using the Docker Image Can Be Customized option you can specify the docker image to use for example
+   * foo/bar:version1 which is appended at the end of the mesos task name (then limited to 64 chars)
+   *
+   * This is a problem because mesos task names cannot contain "/" as the mesos task name is used in
+   * the directory structure to run the task
+   * It is also a problem for the docker hostname which cannot contain ":" or "/"
+   * See RFC 952
+   *
+   * @param name the task or hostname
+   * @return cleaned up hostname or task name where invalid chars are replaced by the "_" character
+   */
+  private static String replaceInvalidHostnameCharacters(String name) {
+    return name.replaceAll("[^a-zA-Z0-9-]", "-");
+  }
 
-        TaskID taskId = TaskID.newBuilder().setValue(name).build();
+  public synchronized void terminateJenkinsSlave(String name) {
+    LOGGER.info("Terminating jenkins slave " + name);
+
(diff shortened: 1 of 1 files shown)


sbeaulie:
candidate fix for #295 too

eThomasHerrlin:
More RFC compliant: `name.replaceAll("[^a-zA-Z0-9-]", "-");`
"_" is not allowed and "." may cause problems also, safest to remove them.

RFC 952:
[...]alphabet (A-Z), digits (0-9), minus
   sign (-), and period (.).  Note that periods are only allowed when
   they serve to delimit components of "domain style names".[...]
Later RFCs extends length and removes limit of first char being Alpha.
Possibly also need to left strip any "-" after replace.

sbeaulie:
@eThomasHerrlin I have changed the regexp accordingly.

By the way this is still an issue with the latest 0.18.1 release, for instance when setting a custom image to `bitnami/ruby:latest`

I get the following error in the jenkins logs:
```
Status update: task value: "mesos-jenkins-a88b6ab710254e9c97218e932244cead-worker:bitnami/r"
 is in state TASK_ERROR with message ''mesos-jenkins-a88b6ab710254e9c97218e932244cead-worker:bitnami/r' contains invalid characters'
```

I've rebased the fix on top of 0.18.1 and released it for my use, but is there any way to get traction on merging this PR?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -243,10 +243,26 @@ private static String joinPaths(String prefix, String suffix) {
...
+   *
+   * This is a problem because mesos task names cannot contain "/" as the mesos task name is used in
+   * the directory structure to run the task
+   * It is also a problem for the docker hostname which cannot contain ":" or "/"
+   * See RFC 952
+   *
+   * @param name the task or hostname
+   * @return cleaned up hostname or task name where invalid chars are replaced by the "_" character
+   */
+  private static String replaceInvalidHostnameCharacters(String name) {
```
vinodkone: Instead of fixing the task name, why dont we fix the slave name instead? That would be change in one place (MesosUtils.buildNodeName()) which I think is cleaner.

vinodkone: I'll be happy to merge this once comments get resolved. Thanks for fixing the issue and sorry it took so long to get feedback.

sbeaulie: would changing the slave name ripple through to the task name?

vinodkone: yes, they are the same.

jeschkies:
Ping @sbeaulie 🙂

sbeaulie:
opening new PR

PR #310, opened 2018-01-24
[JENKINS-49117] - Stop serializing ReentrantLocks to the disk

oleg-nenashev:
It includes #309 for testing purposes and includes a hotfix for https://issues.jenkins-ci.org/browse/JENKINS-49117 to make the plugin compatible with Jenkins 2.102+

@reviewbybees @jglick

PR #311, opened 2018-03-21
[JENKINS-50303] Using XStream to serialize a JSONObject was never advisable and as of 2.102 is forbidden

jglick:
[JENKINS-50303](https://issues.jenkins-ci.org/browse/JENKINS-50303). Follows up #310.

Does not attempt to load historical `slaveAttributes` settings. They will just need to be reconfigured. Since there are barely over a thousand registered users of this plugin, and this field was in an **Advanced** block to begin with, I figure release notes should suffice.

@reviewbybees esp. @oleg-nenashev @Vlatombe

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,2 +1,2 @@
 // Build the plugin using https://github.com/jenkins-infra/pipeline-library
-buildPlugin()
+buildPlugin(jenkinsVersions: [null, '2.107.1'])
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
@@ -90,6 +90,8 @@ public class MesosSlaveInfo extends AbstractDescribableImpl<MesosSlaveInfo> {
   private final boolean defaultSlave;
   // Slave attributes JSON representation.
   private final String slaveAttributesString;
+  @Deprecated
+  private final JSONObject slaveAttributes;
   private final ContainerInfo containerInfo;
   private final List<URI> additionalURIs;
   private final Mode mode;
@@ -250,6 +252,14 @@ public class MesosSlaveInfo extends AbstractDescribableImpl<MesosSlaveInfo> {
     }
 
     return null;
+  }
+
+  private Object readResolve() {
+      if (slaveAttributes != null) {
+          slaveAttributesString = slaveAttributes.toString();
+          slaveAttributes = null;
+      }
+      return this;
   }
 
   public double getdiskNeeded() {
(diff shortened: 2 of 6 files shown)


Jenkinsfile
```diff
@@ -1,2 +1,2 @@
 // Build the plugin using https://github.com/jenkins-infra/pipeline-library
-buildPlugin()
+buildPlugin(jenkinsVersions: [null, '2.107.1'])
```
jglick: > Loading trusted files from base branch master at 670eb0c62db49dd8921489b8fbec6e74c9c0f969 rather than 777bf3062fad1b205228a1818711ea5779748cd1

I manually ran a full build against 2.107.1 and it passed, so probably good enough.

jglick:
> Shouldn't the `readResolve` be implemented anyway for people migrating to the newer plugin version but staying on the same core < 2.102 version?

Well, that would require restoring the `JSONObject slaveAttributes` field, meaning that people upgrading to JEP-200 would get a warning in the system log and unreadable data. I guess that is a reasonable compromise to allow people on pre-JEP-200 to migrate data.

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -91,7 +89,9 @@ else if(maxExecutorsVal < minExecutorsVal) {
   private final String jnlpArgs;
   private final boolean defaultSlave;
   // Slave attributes JSON representation.
-  private final JSONObject slaveAttributes;
+  private String slaveAttributesString;
+  @Deprecated
+  private JSONObject slaveAttributes;
```
oleg-nenashev: mark as transient and send through readResolve() ?

jglick: It _is_ sent through `readResolve`. It could be marked `transient`, though it makes no difference since we no longer write to it except to null it out.

oleg-nenashev: Better to make it transient then. I hope we eventually implement some JEP-200 static analysis, and such case would appear be on our radar for sure

oleg-nenashev:
Thanks @Vlatombe! Please let me know once it is released, I will update the Wiki pages

Vlatombe:
@oleg-nenashev mesos plugin 0.16 is out

karthikpaidi:
We ar using mesos framework for slaves(Jenkins) so when our capacity reaches 90 percent all of sudden slaves start staying in hung state even after finishing the job. It requires a full restart of jenkins did any one saw this issue. any clue why it's happening or any solution ? (meosos 15.1 and 16.0)

PR #315, opened 2018-04-16
Upstream Several Bug Fixes

gabrielhartmann:


gabrielhartmann:
@Vlatombe @jglick Do you know who the appropriate people are to have a look at these changes?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -18,870 +18,924 @@
...
-        if (resource.getType().equals(Value.Type.SCALAR)) {
-          cpus = resource.getScalar().getValue();
+
+    @VisibleForTesting
+    String getFrameworkId() {
+        if (frameworkId != null) {
+            return frameworkId.getValue();
         } else {
-          LOGGER.severe("Cpus resource was not a scalar: " + resource.getType().toString());
+            return NULL_FRAMEWORK_ID;
```
maselvaraj: Wondering if this line would ever be executed. Is it possible resourceOffers() gets called before registered() ?

src/main/java/org/jenkinsci/plugins/mesos/OfferQueue.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * This class acts as a buffer of Offers from Mesos.  By default it holds a maximum of 100 Offers.
+ */
+public class OfferQueue {
+    private final Logger logger = Logger.getLogger(JenkinsScheduler.class.getName());
+
+    private static final int DEFAULT_CAPACITY = 100;
+    private static final Duration DEFAULT_OFFER_WAIT = Duration.ofSeconds(30);
```
maselvaraj: Can you please make DEFAULT_OFFER_WAIT configurable and default it to a minimum time ? For ex: We have 3000 + Jenkins instances. Each Jenkins holding its offers for 30 seconds may leave many Jenkins instances starving ...

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -18,870 +18,924 @@
...
-        LOGGER.info("Created " + processedRequests + " tasks from " + offers.size() + " offers.");
-      }
-    } else {
-      if (!requests.isEmpty()) {
-        LOGGER.info("Did not match any of the " + offers.size() + " offers (" + requests.size() + " pending requests)");
-      }
+
+    @Override
+    public void registered(SchedulerDriver driver, FrameworkID frameworkId, MasterInfo masterInfo) {
+        LOGGER.info("Framework registered! ID = " + frameworkId.getValue());
```
maselvaraj: Related to line 391: Wondering if Mesos would call this method with 'null' for frameworkId then, frameworkId.getValue() would throw a null pointer exception. It may need a null check ?

gabrielhartmann:
@maselvaraj: What version number would be appropriate here?

PR #320, opened 2018-06-06
Make it possible to disable Metrics healthcheck - close #321

### LINKED ISSUE #321: Make it possible to disable Metrics healthcheck
Hi.

In our use infrastructure, we use the metrics plugin to provide a healthcheck for our instances, and we have tools relying on it to evaluate if the Jenkins instance is up.

We witnessed that sometimes this healthcheck is failing because the Mesos plugin indicates we have unmatched labels (_cf._ [MesosAdministrativeMonitor.java#L23](https://github.com/jenkinsci/mesos-plugin/blob/master/src/main/java/org/jenkinsci/plugins/mesos/MesosAdministrativeMonitor.java#L23)).

Hence, we would like to make the metrics healtcheck exposed by this plugin optional, throuch a checkbox in the `/configure` page.

Is that OK for you ?

We submitted PR #320 that implements this.
It includes 5 unit tests and also fixes the current CI state (by correcting 3 minor Findbugs issues).

Lucas-C:
New checkbox in "advanced" section:

2018-06-06 15_07_46-configure system jenkins

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosAdministrativeMonitor.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosAdministrativeMonitor.java
@@ -3,6 +3,7 @@ package org.jenkinsci.plugins.mesos;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
 import hudson.model.AdministrativeMonitor;
+import jenkins.model.Jenkins;
 import org.apache.mesos.Scheduler;
 
 import java.util.HashSet;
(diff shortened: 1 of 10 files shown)


Lucas-C:
I updated this PR to includes fixes for the 3 findbugs violations

Lucas-C:
@vinodkone could you please review this PR ?

jeschkies:
@Lucas-C, thanks for you patch. Would you mind updating it to `1.x`?

Lucas-C:
Done
Once this is merged, you may want to add this new field to this list on the `master` branch for backward compatibility: https://github.com/jenkinsci/mesos-plugin/blob/master/src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java#L89

PR #323, opened 2018-06-12
do not have a release version on master

jtnord:
releases should be made with the `maven-release-plugin` or other, but release versions must never be set on the tip of a branch, just tags.


Vlatombe:
Fixed

PR #324, opened 2018-06-20
Add radioBlock for USER networking on config.jelly

serhermar-stratio:
The option to put the USER-type network is added in the containerInfo configuration file.


sbeaulie:
seemingly a duplicate of #298

PR #325, opened 2018-06-28
original statement is incorrect

gilzellner:
curling to the url doesn't mean it manage to install anything or run anything.


vinodkone:
"Connect to Mesos" just means the TCP connection is successful, it doesn't imply anything about registration/subscription of the scheduler. So I think the wording is fine? The other log changes look fine, so if you want them in please update the PR and I will be happy to merge it.

gilzellner:
Hi,
I no longer work with mesos and don't feel pushing something that I cannot test is appropriate.
I can close this is you don't wish to merge it, but I will not change it since I have no way to test it now without significant work (create mesos cluster etc).

Issue #327, opened 2018-07-19
Mesos Leader Selection


JC1738:
This plugin is great, thank you!  I had a question about the mesos master:5050 link.  Currently everything works great once I enter one of the leaders, but it will all stop working if a new master is elected leader.  It starts working again once we change the host to the newly elected leader. 

I tried to use master:5050/redirect as the host but that didn’t work (didn’t expect it too, but figured that would be a nice feature?)

Thank you,
Jim

geoffroyjabouley:
Hi
please see here: https://github.com/jenkinsci/mesos-plugin/issues/306
You have to use the zk:// url format to properly configure the plugin with a mesos cluster.

JC1738:
@geoffroyjabouley Thanks, I tried to input as suggested and I it doesn't seem to work.  I see the following error in the jenkins logs.

Here is what is in plugin:

`zk://mesos1-qa-XXX.com:5050,mesos2-qa-XXX.com:5050,mesos3-qa-XXX.com:5050`

Here is logs:

`Failed to create a master detector for 'zk://mesos1-qa-XXX.com:5050,mesos2-qa-XXX.com:5050,mesos3-qa-XXX.com:5050': Expecting a (chroot) path for ZooKeeper ('/' is not supported)`

geoffroyjabouley:
I have created a PR: #332
For now it is just documented in README.md, but probably updating the jelly file shall also be done. If i find some time i'll complete the PR

Issue #306, opened 2017-12-11
Jenkins can only launch slaves when talking to current mesos master 


mwmanley:
I have redundant mesos masters that elect through ZooKeeper, as is custom.  Other applications like marathon seem to be able to handle finding the master on their own, but Jenkins seems to only need to be connected to a single master, but has to be connected to the elected master.  So if this master moves due to maintenance or whatever, then Jenkins loses connectivity and cannot launch any slaves.

Has anyone else seen this behaviour?  We're running 1.2 here.  If there is something grossly missing or if I have the ability to enumerate all my masters and it can do the right thing from there, please let me know.  Otherwise, this seems like a bug.

Thanks much!

aleksejlopasov:
+1
Just wanted to create a new issue but found this one first.
The same behavior - Jenkins connects to current active master only. If master role jumps to another server, then Jenkins loses connectivity and no slaves created any more until Jenkins restarts

To solve this I have the temporary solution - another service that resolves name 'leader.mesos' and every 20 seconds, if ip address has changed, then restart signal to Jenkins is sent
(Of course mesos-dns service has be deployed to resolve 'leader.mesos' name)

Please implement feature of finding new master automatically (like marathon/chronos frameworks)

P.S. I tried to create three mesos clouds in Jenkins each configured to its own mesos-master (Only one had a chance to register in mesos due to only one mesos-master is active at the moment). The result is the same

sumeshkanayi:
We have the same problem too

mwmanley:
Once I set it up to use the mesos zookeeper nodes, then it followed the master correctly.  It uses the same syntax as mesos itself with the zk://whatever:2181/path. 

I forgot to close this!

mwmanley:
That's putting the zk paths where it currently asks for the mesos master, I believe.  We moved to k8s from mesos, so I no longer have the syntax handy.

PR #332, opened 2018-09-18
[Doc] Mesos master url configuration with Zookeeper

geoffroyjabouley:
More details reg. Mesos master url configuration of the Mesos plugin (address Zookeeper syntax)

Question has been asked several times in Issues, maybe it is better to document it:
#327 
#306


geoffroyjabouley:
UI change:
image

Issue #334, opened 2018-10-08
wrong description of parameter 'Decline offer duration'


chrisliu529:
It's said ' In milliseconds.' in current help description, however, according to implementation, it's actually in seconds. Therefore, the intended 10 minutes default value (600000) is actually nearly 7 days..

Surprisingly, this issue can hardly be noticed by users because all offers will be revived when a new build request comes. I won't notice it also if I didn't have a deep code reading to analyze issue https://github.com/jenkinsci/mesos-plugin/issues/328

jeschkies:
Fixed in #344.

Issue #328, opened 2018-07-23
some unexpected exceptions in v0.17


chrisliu529:
Last week I upgraded mesos-plugin from v0.16 to v0.17 but I found the 'Jenkins Scheduler' framework stopped working from time to time so I had to downgrade to v0.16.

When the framework stopped working, it didn't accept or decline offers from some agents and "outstanding offers" from these agents could be found in the mesos web UI. If offers from all agents were stuck then no job could be built. In this case, I have to restart mesos master to get it fixed temporarily but this issue may appear again in one day or two.

I checked the logs and found a lot of error logs like 'SEVERE: Offer processing thread failed with exception: java.lang.NullPointerException' which I've never seen in v0.16.

Hope it helps and please let me know if more info is required.
Thanks.

chrisliu529:
The same issue still remains in latest v0.18.1. I found some interesting logs: 
--------------------------------------------------------------------------------
Sep 26, 2018 9:28:43 PM org.jenkinsci.plugins.mesos.JenkinsScheduler processOffers
INFO: Processing 7 offers.
Sep 26, 2018 9:28:43 PM org.jenkinsci.plugins.mesos.JenkinsScheduler$2 run
INFO: Started offer processing thread: mesos-offer-processor-167797f6-bce1-4605-b0d1-dd397bf10201-0007
Sep 26, 2018 9:28:43 PM org.jenkinsci.plugins.mesos.JenkinsScheduler$2 run
SEVERE: Offer processing thread failed with exception: java.lang.NullPointerException
	at org.jenkinsci.plugins.mesos.JenkinsScheduler.declineOffer(JenkinsScheduler.java:310)
	at org.jenkinsci.plugins.mesos.JenkinsScheduler.declineShort(JenkinsScheduler.java:304)
	at org.jenkinsci.plugins.mesos.JenkinsScheduler.processOffers(JenkinsScheduler.java:361)
	at org.jenkinsci.plugins.mesos.JenkinsScheduler.access$500(JenkinsScheduler.java:52)
	at org.jenkinsci.plugins.mesos.JenkinsScheduler$2.run(JenkinsScheduler.java:427)
	at java.lang.Thread.run(Thread.java:748)

Sep 26, 2018 9:28:53 PM org.jenkinsci.plugins.mesos.MesosComputerLauncher launch
INFO: Sending a request to start jenkins slave mesos-jenkins-f971d36ea18c4d7e8faa3a86405d6336-docker
[...]

PR #338, opened 2018-11-28
several fixes and changes on stability, usability and readablity

chrisliu529:
As a heavy user of jenkins mesos framework, I found this plugin become very unstable since v0.17 because of the new multi-threading model. If I enable 'On-demand framework registration', the jenkins framework will crash from time to time and freeze all mesos offers. With bug reported but no response for months, I have to create my own fork and fix on it. Now I'm happy to see v0.18.2 is in plan and hopefully my fixes would help a little.

Key changes:
* fix for #328, now it's safe to enable 'On-demand framework registration'
* make sure any received offer is answered so the offered resource won't be frozen
* fix concurrent read/write on requests queue, which drastically improved the stability

Other changes:
* fix for #334, providing a correct description and implementation for parameter 'Decline offer duration'
* shorten node id to improve the readability of computer names
* add a one-click build script 'dkbd.sh' powered by docker-compose
* a few style adjustments

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -190,6 +190,14 @@
                   </systemProperties>
               </configuration>
           </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.22.1</version>
+                <configuration>
+                    <useSystemClassLoader>false</useSystemClassLoader>
+                </configuration>
+            </plugin>
       </plugins>
   </build>
 </project>
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -41,7 +41,20 @@ import org.apache.mesos.Protos.Volume.Mode;
 import org.apache.mesos.Scheduler;
 import org.apache.mesos.SchedulerDriver;
 
-import java.util.*;
+import java.util.Queue;
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.HashSet;
+import java.util.TreeSet;
+import java.util.Map;
+import java.util.HashMap;
+import java.util.List;
+import java.util.LinkedList;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.Date;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
@@ -90,7 +103,7 @@ public class JenkinsScheduler implements Scheduler {
 
     private static final Object IGNORE = new Object();
 
-    private static final OfferQueue offerQueue = new OfferQueue();
+    private final OfferQueue offerQueue = new OfferQueue();
     private Thread offerProcessingThread = null;
     private volatile FrameworkID frameworkId;
 
@@ -423,7 +436,7 @@ public class JenkinsScheduler implements Scheduler {
             public void run() {
                 LOGGER.info("Started offer processing thread: " + threadName);
                 try {
-                    while (true) {
+                    while (isRunning()) {
                         processOffers();
                     }
                 } catch (Throwable t) {
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -159,6 +159,7 @@ public class JenkinsScheduler implements Scheduler {
         } else {
             driver = new MesosSchedulerDriver(JenkinsScheduler.this, framework, mesosCloud.getMaster());
         }
+
         // Start the framework.
         Thread frameworkThread = new Thread(new Runnable() {
             @Override
@@ -173,33 +174,33 @@ public class JenkinsScheduler implements Scheduler {
                 } catch(RuntimeException e) {
                     LOGGER.log(Level.SEVERE, "Caught a RuntimeException", e);
                 } finally {
-                    SUPERVISOR_LOCK.lock();
-                    if (driver != null) {
-                        driver.abort();
-                    }
-                    driver = null;
                     running = false;
-                    SUPERVISOR_LOCK.unlock();
+                    SUPERVISOR_LOCK.lock();
+                    try {
+                        if (driver != null) {
+                            driver.abort();
+                        }
+                        driver = null;
+                    } finally {
+                        SUPERVISOR_LOCK.unlock();
+                    }
                 }
             }
-        });
-
-        frameworkThread.setName("mesos-framework-thread");
+        }, String.format("mesos-framework-thread-%d", startedTime));
         frameworkThread.setDaemon(true);
-
         frameworkThread.start();
     }
 
     public synchronized void stop() {
+        running = false;
+        SUPERVISOR_LOCK.lock();
         try {
-            SUPERVISOR_LOCK.lock();
             if (driver != null) {
                 LOGGER.info("Stopping Mesos driver.");
                 driver.stop();
             } else {
                 LOGGER.warning("Unable to stop Mesos driver:  driver is null.");
             }
-            running = false;
         } finally {
             SUPERVISOR_LOCK.unlock();
         }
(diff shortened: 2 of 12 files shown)


src/main/java/org/jenkinsci/plugins/mesos/OfferQueue.java
```diff
@@ -22,7 +22,7 @@
     private final Logger logger = Logger.getLogger(JenkinsScheduler.class.getName());
 
     private static final int DEFAULT_CAPACITY = 100;
-    private static final Duration DEFAULT_OFFER_WAIT = Duration.ofSeconds(30);
+    private static final Duration DEFAULT_OFFER_WAIT = Duration.ofSeconds(600);
```
sbeaulie: Why change this value here?

sbeaulie:
The more I look at this PR I think it is the right fix for some of our instability issues (NPE when processing offers). How has it been running for you @chrisliu529 ?

src/main/java/org/jenkinsci/plugins/mesos/OfferQueue.java
chrisliu529: Here I just enlarged the race-condition window (as explained in issue #328 ) to make my tests easier and the modified model more convincing.

chrisliu529:
> The more I look at this PR I think it is the right fix for some of our instability issues (NPE when processing offers). How has it been running for you @chrisliu529 ?

Yes. Having been running in production for 2 months and its stability proved, I issued this PR. And now 5 more months passed, it works as solid as rocks :)

sbeaulie:
Thank you. One symptom we saw without this fix, when the queue was frozen are 'outstanding offers' on the mesos cluster `mesos.master.outstanding_offers`. This metric would pop up and jenkins would be stuck processing the same offer in loop.

I did build this PR but I had to remove one commit:
https://github.com/jenkinsci/mesos-plugin/pull/338/commits/7dca44219e617c492661a9be7d3e518eb8a62033

It might be an existing hidden bug that this commit makes surface but we would not be able to launch docker images that are 'customized'. The mesos task would get staged and try to start and get the jpnl link from jenkins `http://<jenkins_master>/computer/<node_id-label>/slave-agent.jnlp?encrypt=true` and fail with a 404, probably because of the presence of unencoded chars in the url.

chrisliu529:
> Thank you. One symptom we saw without this fix, when the queue was frozen are 'outstanding offers' on the mesos cluster `mesos.master.outstanding_offers`. This metric would pop up and jenkins would be stuck processing the same offer in loop.
> 
> I did build this PR but I had to remove one commit:
> [7dca442](https://github.com/jenkinsci/mesos-plugin/commit/7dca44219e617c492661a9be7d3e518eb8a62033)
> 
> It might be an existing hidden bug that this commit makes surface but we would not be able to launch docker images that are 'customized'. The mesos task would get staged and try to start and get the jpnl link from jenkins `http://<jenkins_master>/computer/<node_id-label>/slave-agent.jnlp?encrypt=true` and fail with a 404, probably because of the presence of unencoded chars in the url.

The 404 issue is not caused by commit 7dca442. It's because the jenkins web UI is not timely sync with the backend data, especially when a container with a short timeout (e.g 1 minute) configured.

nemacysts:
@maselvaraj (assuming you're still a maintainer): any blockers on getting this reviewed?

src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
```diff
@@ -58,8 +58,8 @@
             <st:nbsp/>${%No}
         </f:entry>
 
-        <f:entry title="${%Decline offer duration}" field="declineOfferDuration" description="${%Decline offers for specified amount of time in case no slave is queued. The plugin asks for new offers in case a new slave needs to be started. In milliseconds.}">
-          <f:number clazz="required positive-number" min="1000" steps="1000" default="600000"/>
+        <f:entry title="${%Decline offer duration}" field="declineOfferDuration" description="${%Decline offers for specified amount of time in case no slave is queued. The plugin asks for new offers in case a new slave needs to be started. In seconds.}">
```
nemacysts: Could changing the unit here without a data migration cause issues for some users?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -82,15 +94,15 @@
 
     private static final Logger LOGGER = Logger.getLogger(JenkinsScheduler.class.getName());
 
-    public static final Lock SUPERVISOR_LOCK = new ReentrantLock();
+    public static final ReentrantLock SUPERVISOR_LOCK = new ReentrantLock();
```
nemacysts: Is this change necessary? Seems like we don't need to go from an interface to a more concrete version in the definition

sbeaulie:
> The 404 issue is not caused by commit [7dca442](https://github.com/jenkinsci/mesos-plugin/commit/7dca44219e617c492661a9be7d3e518eb8a62033). It's because the jenkins web UI is not timely sync with the backend data, especially when a container with a short timeout (e.g 1 minute) configured.

This issue is reproducible. I could see the computer in the jenkins UI, but the mesos task was never able to connect to it. Also these were for brand new computers, and not a reuse of them. All the jobs using a `custom docker image` with the mesos label would not start, and mesos would continually try in a loop to start a task and the task would fail. The list of offline computers/nodes would grow. Without the commit it works.

What is the timeout you are referring to? Our idle timeout value is 3 minutes

chrisliu529:
> > The 404 issue is not caused by commit [7dca442](https://github.com/jenkinsci/mesos-plugin/commit/7dca44219e617c492661a9be7d3e518eb8a62033). It's because the jenkins web UI is not timely sync with the backend data, especially when a container with a short timeout (e.g 1 minute) configured.
> 
> This issue is reproducible. I could see the computer in the jenkins UI, but the mesos task was never able to connect to it. Also these were for brand new computers, and not a reuse of them. All the jobs using a `custom docker image` with the mesos label would not start, and mesos would continually try in a loop to start a task and the task would fail. The list of offline computers/nodes would grow. Without the commit it works.
> 
> What is the timeout you are referring to? Our idle timeout value is 3 minutes

I'm afraid we are not talking about the same issue. The 404 issue is because the container is already recycled when its idle timeout is reached but the UI has not been synced yet so there are some invalid links.

Concerning the connection issue, there could be several possible reasons but currently I have no idea how it could link to the shortened computer names. Anyway it works well in my production. Maybe you can find some clues in the 'Jenkins System Log' by adding a logger 'org.jenkinsci.plugins.mesos' with 'fine' level?

src/main/resources/org/jenkinsci/plugins/mesos/MesosCloud/config.jelly
chrisliu529: Data migration is not required here because the change removed some restrictions instead of adding.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
chrisliu529: It's for clearance, indicating this lock is preventing reentrance only.

jeschkies:
👋 these changes look good to me on first sight. We've recently rewrote the plugin to be completely async. Would you mind trying that: https://github.com/mesosphere/jenkins-mesos-plugin? I hope to take ownership of this repo with the new plugin once it's stable enough.

pom.xml
```diff
@@ -190,6 +190,14 @@
                   </systemProperties>
               </configuration>
           </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
```
vinodkone: why this change?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -423,7 +455,7 @@ void startProcessing() {
             public void run() {
                 LOGGER.info("Started offer processing thread: " + threadName);
                 try {
-                    while (true) {
+                    while (isRunning()) {
```
vinodkone: What's the guarantee that between the `isRunning()` check and `processOffers()` below that the `driver` wont be set to null?

PR #344, opened 2019-07-04
change timeunit for decline-offer-duration parameter in config.jelly from ms to seconds

erikzquby:
This is a fix for:
https://github.com/jenkinsci/mesos-plugin/issues/334

Renamed ms (milliseconds) to s (seconds), and divided number parameters by 1000.

The values from the config.jelly get passed to org.apache.mesos.Protos.Filters.Builder#setRefuseSeconds in org.jenkinsci.plugins.mesos.JenkinsScheduler#declineOffer, this method accepts seconds.


erikzquby:
This is a duplicate of a small part of https://github.com/jenkinsci/mesos-plugin/pull/338

pwlmesos:
any chance of approving PR#338 as well?

PR #345, opened 2019-07-31
Fix for logging mesos slave hostname

sathiya-mit:
If we run Jenkins slave in docker container, we always get hostname as below
Mesos slave(hostname): 172.18.0.1

I have fixed the same. Please review it.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/listener/MesosRunListener.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/listener/MesosRunListener.java
@@ -1,5 +1,6 @@
 package org.jenkinsci.plugins.mesos.listener;
 
+import hudson.EnvVars;
 import hudson.Extension;
 import hudson.model.AbstractBuild;
 import hudson.model.Computer;
@@ -19,7 +20,9 @@ import org.jenkinsci.plugins.mesos.MesosSlave;
 public class MesosRunListener extends RunListener<Run> {
 
   private static final Logger LOGGER = Logger.getLogger(MesosRunListener.class.getName());
-  
+  private static final String HOSTNAME = "HOSTNAME";
+  private static final String DOCKER = "DOCKER";
+
   public MesosRunListener() {
     
   }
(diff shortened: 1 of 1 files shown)


sathiya-mit:
> LGTM. Have you tested this with and without docker container based Jenkins agent?

I have taken care. In both the cases it should work now. Please check.

src/main/java/org/jenkinsci/plugins/mesos/listener/MesosRunListener.java
```diff
@@ -73,4 +82,11 @@ private final Node getCurrentNode() {
     return null;
   }
 
+  private boolean isBuildRunningInsideContainer(EnvVars environment) {
+    if(environment.get(MESOS_CONTAINER_NAME) != null) {
```
vinodkone: In Mesos, a container can be run using Mesos Containerizer or Docker containerizer. AFAICT, this env variable is only set when using Docker containerizer. So we should rename this method name to `isBuildRunningwithDockerContainerizer` but if we take my above suggestion, we can kill this method altogether.

src/main/java/org/jenkinsci/plugins/mesos/listener/MesosRunListener.java
```diff
@@ -44,7 +47,13 @@ public void onStarted(Run r, TaskListener listener) {
         try {
           Computer computer = node.toComputer();
           if (computer != null) {
-            String hostname = computer.getHostName();
+            EnvVars environment = computer.getEnvironment();
+            String hostname;
+            if(isBuildRunningInsideContainer(environment)) {
+              hostname = environment.get(HOSTNAME);
+            }else {
+              hostname = node.toComputer().getHostName();
```
vinodkone: hostname = computer.getHostName();

src/main/java/org/jenkinsci/plugins/mesos/listener/MesosRunListener.java
```diff
@@ -44,7 +47,13 @@ public void onStarted(Run r, TaskListener listener) {
         try {
           Computer computer = node.toComputer();
           if (computer != null) {
-            String hostname = computer.getHostName();
+            EnvVars environment = computer.getEnvironment();
+            String hostname;
+            if(isBuildRunningInsideContainer(environment)) {
+              hostname = environment.get(HOSTNAME);
+            }else {
```
vinodkone: space after "}"

src/main/java/org/jenkinsci/plugins/mesos/listener/MesosRunListener.java
```diff
@@ -44,7 +47,13 @@ public void onStarted(Run r, TaskListener listener) {
         try {
           Computer computer = node.toComputer();
           if (computer != null) {
-            String hostname = computer.getHostName();
+            EnvVars environment = computer.getEnvironment();
+            String hostname;
+            if(isBuildRunningInsideContainer(environment)) {
```
vinodkone: space between "if" and "(".

vinodkone: Instead of depending on the env variable can you do `node.getSlaveInfo().getContainerInfo().type() == DOCKER`?

sathiya-mit:
@vinodkone Thanks for your valuable feedback. I have addressed your review comments. Please do the needful.

sathiya-mit:
@vinodkone If you don't have any other review comments, could you please merge this PR ?

sathiya-mit:
@vinodkone Thank You !!!

PR #340, opened 2019-03-10
Support container images with mesos containerizer 

kamaradclimber:
This PR introduces support to launch jenkins slave with mesos containerizer within a docker image.
This allows to benefit from mesos containerizer while getting filesystem isolation.

This is my first jenkins plugin patch and I would really appreciate feedback/advices (on jelly files especially).

PR also build against latest mesos version although it is not strictly necessary (I can remove this commit if needed)

PR #347, opened 2019-08-08
support UCR runtime

kvish:


src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -927,6 +930,20 @@ private void getContainerInfoBuilder(Offer offer, Request request, String slaveN
 
                 containerInfoBuilder.setDocker(dockerInfoBuilder);
                 break;
+            case MESOS:
```
vinodkone: Dont' we need to expose this option in the plugin configuration?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -1007,6 +1024,7 @@ String generateJenkinsCommand2Run(int jvmMem,String jvmArgString,String jnlpArgS
 
         } else {
             LOGGER.fine("About to use default shell ....");
+            commandBuilder.setShell(true);
```
vinodkone: What is this for?

vinodkone:
See #340

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
kvish: removed

src/main/resources/org/jenkinsci/plugins/mesos/MesosSlaveInfo/ContainerInfo/config.jelly
```diff
@@ -38,7 +40,6 @@
                 </f:repeatableProperty>
             </f:entry>
         </f:radioBlock>
-        <f:radioBlock id="networking.USER" name="networking" title="${%User}" value="USER" inline="true" checked="${instance.networking == null || instance.networking == 'USER'}" />
```
vinodkone: Why was this deleted? Backwards compat?

src/main/resources/org/jenkinsci/plugins/mesos/MesosSlaveInfo/ContainerInfo/config.jelly
```diff
@@ -1,10 +1,10 @@
-<?jelly escape-by-default='true'?>
+?jelly escape-by-default='true'?>
```
vinodkone: open "<" missing?

src/main/resources/org/jenkinsci/plugins/mesos/MesosSlaveInfo/ContainerInfo/config.jelly
```diff
@@ -1,10 +1,10 @@
...
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:c="/lib/credentials">
     <f:entry title="${%Container Type}" field="type">
-        <f:radioBlock name="type" title="${%Docker}" value="DOCKER" inline="true" checked="true">
-            <f:entry title="${%Docker Image}" field="dockerImage" description="If using Docker, specify the docker image.">
+            <f:entry title="${%Docker Image}" field="dockerImage" description="Specify the docker image.">
```
vinodkone: Is this change backwards compatible? If someone upgrades the plugin and used previous config, sounds like it would fail to start? I think we probably want to just add another entry for docker image name inside the new radioblock for compat.

vinodkone: As discussed offline, we need to fix this for backwards compat.

kvish: tested this and we should be good
- downgrades are not supported in jenkins so we are only concerned about uopgrade path
- When upgrading the containerType will be set to Docker correctly on the agent spec and the user can override this by selecting UCR

PR #348, opened 2019-08-12
Fixed redundant suppress.

vinodkone:
No need to suppress when there were no offers received.


src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -394,7 +394,7 @@ private void processOffers() {
 
         // Note that even if the suppress call is dropped on the way to the master it is ok
 	// because we will do this check again when processing a subsequent offer.
-        if (!pendingWork()) {
+        if (offers.size() > 0 && !pendingWork()) {
             unmatchedLabels.clear();
```
kvish: nit: can we add a log line here when we suppress

vinodkone: We log in `suppressOffers()` already.

PR #349, opened 2019-08-13
Enhance metrics

kvish:


src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -773,6 +793,11 @@ private void createMesosTask(Offer offer, Request request) {
 
         Metrics.metricRegistry().counter("mesos.scheduler.operation.launch").inc(tasks.size());
         Filters filters = Filters.newBuilder().setRefuseSeconds(1).build();
+
+
+        request.request.mesosSlave.provisionedToMesos();
```
kaiwalyajoshi: (general question) does this have side-effects?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -773,6 +793,11 @@ private void createMesosTask(Offer offer, Request request) {
...
+
+        LOGGER.info(String.format("Slave %s now being provisioned by Mesos", request.request.mesosSlave.getUuid()));
```
kaiwalyajoshi: I'm getting a build error as `request.request.mesosSlave` is an unkown symbol.
```suggestion
```

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
kaiwalyajoshi: I'm getting a build error as request.request.mesosSlave is an unkown symbol.
```suggestion
```

PR #351, opened 2019-08-13
Add Single Use property to MesosSlave (#16)

kvish:
* Add Single Use property to MesosSlave

Single-use agents are marked offline and then still wait for
idle timeout before being deleted. This commit adds a new
property to MesosSlave to mark single-use agents. This is
read during the retention check and used to mark a cluster
for deletion.

* Order matters

* Add locks around new property

* Add test for single use and retention policy

* Sync pending delete as well

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosSingleUseSlave.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosSingleUseSlave.java
@@ -57,6 +57,22 @@ public class MesosSingleUseSlave extends SimpleBuildWrapper {
     @Override
     public void setUp(Context context, Run<?, ?> build, FilePath workspace, Launcher launcher, TaskListener listener, EnvVars initialEnvironment) throws IOException, InterruptedException {
         context.setDisposer(new MesosSingleUseSlaveDisposer());
+
+        Computer computer = workspace.toComputer();
+        if (computer != null) {
+            if (MesosComputer.class.isInstance(computer)) {
+                String msg = "Marking " + computer.getName() + " as single-use.";
+                LOGGER.warning(msg);
+                listener.getLogger().println(msg);
+
+                MesosSlave mesosSlave = (MesosSlave) computer.getNode();
+                if (mesosSlave != null) {
+                    mesosSlave.setSingleUse(true);
+                }
+            } else {
+                listener.getLogger().println("Not able to set single-use slave, this is a " + computer.getClass());
+            }
+        }
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/mesos/MesosSingleUseSlave.java
```diff
@@ -57,6 +57,22 @@ public MesosSingleUseSlave() {
     @Override
     public void setUp(Context context, Run<?, ?> build, FilePath workspace, Launcher launcher, TaskListener listener, EnvVars initialEnvironment) throws IOException, InterruptedException {
         context.setDisposer(new MesosSingleUseSlaveDisposer());
+
+        Computer computer = workspace.toComputer();
+        if (computer != null) {
+            if (MesosComputer.class.isInstance(computer)) {
+                String msg = "Marking " + computer.getName() + " as single-use.";
+                LOGGER.warning(msg);
```
vinodkone: not sure why this is a `LOGGER.warning`. this could just be `Logger.info` since this is intentional?

src/main/java/org/jenkinsci/plugins/mesos/MesosSlave.java
```diff
@@ -33,14 +33,21 @@
...
+
   private final String cloudName;
   private transient MesosCloud cloud;
   private final MesosSlaveInfo slaveInfo;
   private final int idleTerminationMinutes;
   private final double cpus;
   private final int mem;
   private final double diskNeeded;
-  private transient final Timer.Context provisioningContext;
+  private final UUID uuid = UUID.randomUUID();
```
vinodkone: these are already part of #349 . they shouldn't show up here?

src/main/java/org/jenkinsci/plugins/mesos/MesosSlave.java
```diff
@@ -65,7 +72,10 @@ public MesosSlave(MesosCloud cloud, String name, int numExecutors, MesosSlaveInf
     this.cpus = slaveInfo.getSlaveCpus() + (numExecutors * slaveInfo.getExecutorCpus());
     this.mem = slaveInfo.getSlaveMem() + (numExecutors * slaveInfo.getExecutorMem());
     this.diskNeeded = slaveInfo.getdiskNeeded();
-    this.provisioningContext = provisioningContext;
+    this.provisionToReady = provisionToReadyContext;
```
vinodkone: ditto. see above.

Issue #15, opened 2014-02-04
Terminate jenkins slave task if it doesn't become online after TIMEOUT


tarnfeld:
I've noticed that when there are a lot of builds in the queue, the scheduler will continue to launch more Jenkins slave tasks, regardless of the available resources in the Mesos Cluster (quite rightly). These tasks sit in the mesos queue until the other slave tasks die as there are no available resources.

In my opinion this doesn't make sense, because the other Jenkins slave tasks _aren't_ going to die (due to being idle) when the jobs running on them finish, they're just going to pick up the remaining jobs in the Jenkins build queue.

When all jobs are finished, and the running slave tasks begin to die because they're idle, the tasks in the queue will launch and consume resources that aren't needed – and will subsequently die after the idle timeout.

---
- Would it make sense for there to be a timeout on the tasks, if the slave doesn't come up within this period the jenkins scheduler will kill the task?
- Is there a way of asking Mesos for the tasks currently waiting for available resources, perhaps the scheduler could not launch more tasks if there's one in the queue.

Correct me if i'm wrong, I might be missing something...

PR #352, opened 2019-08-13
Improve logging of offer processing and slave configuration (#15)

kvish:


src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -773,6 +780,14 @@ private void createMesosTask(Offer offer, Request request) {
 
         Metrics.metricRegistry().counter("mesos.scheduler.operation.launch").inc(tasks.size());
         Filters filters = Filters.newBuilder().setRefuseSeconds(1).build();
+
+        for (TaskInfo taskInfo : tasks) {
+            LOGGER.info(String.format("Launching TaskInfo: %s", TextFormat.shortDebugString(taskInfo)));
+        }
+
+        request.request.mesosSlave.provisionedToMesos();
```
vinodkone: this should be part of #349 .

src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -1,24 +1,5 @@
 package org.jenkinsci.plugins.mesos;
```
vinodkone: This diff is hard to read because it's mixing formatting changes with logic changes. But I trust that @kaiwalyajoshi reviewed it properly.

PR #353, opened 2019-08-14
Enhance metrics (#14)

kvish:
* Track per agent offer metrics

* Add better timing metrics

* Adding logging to verify actual timeframe of executor launch

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -297,6 +297,11 @@ public class JenkinsScheduler implements Scheduler {
 
     private void declineShort(Offer offer) {
         Metrics.metricRegistry().meter("mesos.scheduler.decline.short").mark();
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -331,13 +336,22 @@ private void processOffers() {
...
+            }
+
             final Timer.Context offerContext = Metrics.metricRegistry().timer("mesos.scheduler.offer.processing.time").time();
             try {
                 if (!pendingWork()) {
                     // Decline offer for a longer period if no slave is waiting to get spawned.
                     // This prevents unnecessarily getting offers every few seconds and causing
                     // starvation when running a lot of frameworks.
                     Metrics.metricRegistry().meter("mesos.scheduler.decline.long").mark();
+                    LOGGER.info("No slave in queue.");
```
vinodkone: I would delete this log line. Logging this for every offer seems excessive.

PR #359, opened 2019-10-04
Migrate Dind flag.

jeschkies:
Summary:
This resolves the `isDind` flag on the container info to enable a config
migration from 0.x to 1.1.

JIRA issues: DCOS_OSS-5586


src/main/java/org/jenkinsci/plugins/mesos/MesosSlaveInfo.java
```diff
@@ -460,7 +460,7 @@ public String getDisplayName() {
         private final boolean dockerPrivilegedMode;
         private final boolean dockerForcePullImage;
         private final boolean dockerImageCustomizable;
-        private final boolean isDind;
```
timcharper: Are we sure that this is broken? From my understanding, "null" primitive values default to whatever value is specified by 0; For boolean, this is value `false.`

jeschkies: I thought the same. @kaiwalyajoshi wanted to test this.

PR #367, opened 2019-11-06
Replace invalid hostname characters for task names and docker hostname

### LINKED ISSUE #251: Docker slaves launched with invalid hostname when label contains underscores

sbeaulie:
Before this change, if you used the mesos plugin option 'Docker Image Can Be Customized'
and the resulting task name was less than 64 characters it could happen that the docker
image name contained invalid characters both for the task name and the docker hostname.
For example a custom label:foo/bar:latest would fail as a task name because / is invalid
in a task name since it breaks the directory structure. It would also fail as a docker
hostname since it needs to comply with RFC 1123 hostnames

Original commit was to fix #251
Making the change to the buildNodeName() makes it consistent across the code

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosUtils.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosUtils.java
@@ -5,13 +5,22 @@ import java.util.UUID;
 
 public final class MesosUtils {
     private static final int MAX_HOSTNAME_LENGTH = 63; // Guard against LONG hostnames - RFC-1034
+    /** The buildNodeName may contain the jenkins label, and is subsequently
+     * used to name the mesos task and docker hostname.
+     * A mesos task cannot contain "/"
+     * A docker hostname may not contain ":" or "/"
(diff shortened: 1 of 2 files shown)


sbeaulie:
Disclaimer: I have have not tested this as our custom `jar` runs on a different branch based on #302 that I just closed.

PR #370, opened 2019-11-21
Define fault domain filter.

jeschkies:
Summary:
This patch introduces the option to limit Jenkins Mesos nodes to
specific fault domains, ie regions and zones. The filter is a simple
equality matcher for now.

JIRA issues: DCOS-61122


src/main/java/org/jenkinsci/plugins/mesos/api/LaunchCommandBuilder.java
```diff
@@ -174,4 +183,12 @@ private URL buildJnlpUrl() throws MalformedURLException {
         .add(jenkinsAgentFetchUri)
         .build();
   }
+
+  private static DomainFilter ANY_DOMAIN =
```
kaiwalyajoshi: I would suggest having this available from USI rather than defining it here, that way any framework can pick between `AnyDomainFilter`, `HomeRegionFilter`, `RegionFilter` etc... and we build up a library of those ready for use, in a "batteries included" sort of manner.

jeschkies: Good point. There is a follow up at https://github.com/mesosphere/usi/pull/127.

PR #371, opened 2019-11-22
Define agent shell command style.

jeschkies:
Summary:
When the Jenkins agent is launched on a Windows node the command differs slightly
from the Linux command. This patch introduces a command style which defaults to Linux.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosAgentSpecTemplate.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosAgentSpecTemplate.java
@@ -315,27 +315,6 @@ public class MesosAgentSpecTemplate extends AbstractDescribableImpl<MesosAgentSp
         load();
       }
     }
-
-    /**
-     * Migrate older fields from 0.x and 1.x of the plugin.
-     *
-     * @return The migrated container info.
-     */
-    private Object readResolve() {
-
-      // No custom shell command should be used the shell command field must be null.
-      if (!this.useCustomDockerCommandShell) {
-        return new ContainerInfo(
-            this.type,
-            this.dockerImage,
-            this.isDind,
-            this.dockerPrivilegedMode,
-            this.dockerForcePullImage,
-            this.volumes);
-      } else {
-        return this;
-      }
-    }
   }
 
   public static class Volume extends AbstractDescribableImpl<Volume> {
--- a/src/main/java/org/jenkinsci/plugins/mesos/MesosAgentSpecTemplate.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/MesosAgentSpecTemplate.java
@@ -48,6 +48,7 @@ public class MesosAgentSpecTemplate extends AbstractDescribableImpl<MesosAgentSp
   private final int maxExecutors;
   private final String jnlpArgs;
   private final List<MesosSlaveInfo.URI> additionalURIs;
+  private final LaunchCommandBuilder.AgentCommandStyle agentCommandStyle;
   private final ContainerInfo containerInfo;
   private final DomainFilterImpl domainInfoFilter;
 
@@ -64,6 +65,7 @@ public class MesosAgentSpecTemplate extends AbstractDescribableImpl<MesosAgentSp
       String jnlpArgs,
       List<MesosSlaveInfo.URI> additionalURIs,
       ContainerInfo containerInfo,
+      LaunchCommandBuilder.AgentCommandStyle agentCommandStyle,
       DomainFilterImpl domainInfoFilter) {
     this.label = label;
     this.labelSet = Label.parse(label);
@@ -79,6 +81,7 @@ public class MesosAgentSpecTemplate extends AbstractDescribableImpl<MesosAgentSp
     this.additionalURIs = (additionalURIs != null) ? additionalURIs : Collections.emptyList();
     this.containerInfo = containerInfo;
     this.domainInfoFilter = domainInfoFilter;
+    this.agentCommandStyle = agentCommandStyle;
     validate();
   }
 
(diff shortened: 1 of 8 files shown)


src/main/resources/org/jenkinsci/plugins/mesos/MesosAgentSpecTemplate/ContainerInfo/config.jelly
```diff
@@ -15,6 +15,16 @@
         <f:radioBlock name="type" title="${%Mesos}" value="MESOS" inline="true" checked="${instance.type == 'MESOS'}"></f:radioBlock>
     </f:entry>
 
+    <f:block>
```
jeschkies: This is somewhat ugly but `optioanlProperty` did not work on a `String`. Any help is welcome.

src/main/java/org/jenkinsci/plugins/mesos/api/LaunchCommandBuilder.java
```diff
@@ -137,13 +137,19 @@ public LaunchPod build() throws MalformedURLException, URISyntaxException {
...
   /** @return the agent shell command for the Mesos task. */
   private String buildCommand() throws MalformedURLException {
-    return String.format(
-        AGENT_COMMAND_FORMAT,
-        this.xmx,
-        this.jvmArgString,
-        this.jnlpArgString,
-        buildJnlpSecret(),
-        buildJnlpUrl());
+    // Check if command is overriden
```
jeschkies: Please see `SimpleTaskInfoBuilder` in USI. The `shellCommand` will call 
```scala
      taskInfoBuilder.setCommand(
        Mesos.CommandInfo
          .newBuilder()
          .setShell(true)
          .setValue(shellCommand)
```
this is different from [1.x](https://github.com/jenkinsci/mesos-plugin/blob/1.x/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java#L1050) where we have
```java
  commandBuilder.setShell(false);
  commandBuilder.setValue(customShell);
``` 
Kai, do you know why we set shell to `false`?

src/main/resources/org/jenkinsci/plugins/mesos/MesosAgentSpecTemplate/ContainerInfo/help-overrideShellCommand.html
```diff
@@ -0,0 +1,7 @@
+<div>
+  Check if you do not want to use the default command to launch the Jenkins agent. This is useful if
+  you want to start the agent in a Windows container:
+  <code>
+    java -DHUDSON_HOME=jenkins -server -Xmx%dm %s -jar ${MESOS_SANDBOX-.}/agent.jar %s %s -jnlpUrl %s";
```
jeschkies: Kai, do you have a proper example? How would a use know the JNLP URL?

src/main/java/org/jenkinsci/plugins/mesos/api/LaunchCommandBuilder.java
timcharper: shell is set to `true` if we want to pass a shell script to be evaluated by `sh`; `setShell` effectively should've been called `setShellScript(true)`.

jeschkies:
As @kai pointed out I misunderstood the option. We need a new override shell command option.

src/main/resources/org/jenkinsci/plugins/mesos/MesosAgentSpecTemplate/ContainerInfo/help-overrideShellCommand.html
kaiwalyajoshi: The user wouldn't know this and ideally shouldn't configure the JNLP URL, its too much to push on the user and I haven't seen a case where a user requested a custom url.

Right now its derived from here: https://github.com/jenkinsci/mesos-plugin/blob/1.x/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java#L216-L222

I'm in favour of not exposign this URL unless there's concrete use case from the user to change it.

kaiwalyajoshi:
The shell command option is needed for the [dind wrapper](https://github.com/rollbar/docker/blob/master/hack/dind) which customers typically use for their build jobs.

I've found when and where to use the custom-comand shell to be really confusing and its mechanics to be equally archane. If possible with the 2.0 plugin, we should really disambiguate the scenarios in which this custom comand shell is used.

src/main/java/org/jenkinsci/plugins/mesos/MesosAgentSpecTemplate.java
```diff
@@ -307,6 +315,27 @@ public DescriptorImpl() {
         load();
       }
     }
+
+    /**
```
timcharper: Can you help me understand how this relates to the template command? I'm failing to make the connection

src/main/java/org/jenkinsci/plugins/mesos/api/LaunchCommandBuilder.java
```diff
@@ -40,8 +40,15 @@ public LaunchCommandBuilder() {}
...
-  private static final String AGENT_COMMAND_FORMAT =
+  public static enum AgentCommandStyle {
+    Linux,
+    Windows
+  }
+
+  private static final String LINUX_AGENT_COMMAND_TEMPLATE =
       "java -DHUDSON_HOME=jenkins -server -Xmx%dm %s -jar ${MESOS_SANDBOX-.}/agent.jar %s %s -jnlpUrl %s";
+  private static final String WINDOWS_AGENT_COMMAND_TEMPLATE =
+      "java -DHUDSON_HOME=jenkins -server -Xmx%dm %s -jar %MESOS_SANDBOX%/agent.jar %s %s -jnlpUrl %s";
```
kaiwalyajoshi:
```suggestion
      "java -DHUDSON_HOME=jenkins -server -Xmx%dm %s -jar %%MESOS_SANDBOX%%/agent.jar %s %s -jnlpUrl %s";
```
The percentage sign wasn't escaped properly, this leads to failures.

src/main/java/org/jenkinsci/plugins/mesos/MesosAgentSpecTemplate.java
jeschkies: That was a left over from the old diff.

PR #373, opened 2019-11-25
Define agent attribute filters.

kaiwalyajoshi:
Supports only basic string attribute matching for now.


jeschkies:
Close in favour of #374.

PR #377, opened 2019-11-27
Select fault domain filter.

jeschkies:
Summary:
This change enables an operator to choose the home region, any domain or
a custom filter to decide which offers from which fault domain should be
considered.


src/main/java/org/jenkinsci/plugins/mesos/MesosAgentSpecTemplate.java
```diff
@@ -357,31 +357,4 @@ public DescriptorImpl() {
       }
     }
   }
-
-  public static class DomainFilterImpl extends AbstractDescribableImpl<DomainFilterImpl>
```
jeschkies: The other models should be moved to `config.models` as well.

jeschkies:
Here is a tiny demo
faultdomain

PR #384, opened 2019-12-31
Reuse Mesos session.

jeschkies:
Summary:
The plugin would create a new framework ID and a new Mesos connection if the config was
changed. This patch persists the framework ID in the config as a read only parameter.

Furthermore the plugin is now partition and region aware to support Windows fault domains.

We also test the serialization and deserialization.

Jira issues: DCOS_OSS-5831, DCOS_OSS-5817

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,7 +58,7 @@ test {
 }
 
 group = "org.jenkins-ci.plugins"
-version = "2.0-beta-11"
+version = "2.0-beta-12"
 description = "Allows the dynamic launch of Jenkins agent on a Mesos cluster, depending on workload"
 sourceCompatibility = 1.8
 targetCompatibility = 1.8
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ repositories {
 }
 
 ext {
-    usiVersion= '0.1.28-9e5746a-SNAPSHOT'
+    usiVersion= '0.1.28-7b3e510-SNAPSHOT'
     scalaVersion = '2.12'
     akkaVersion = '2.5.19'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -58,7 +58,7 @@ test {
 }
 
 group = "org.jenkins-ci.plugins"
-version = "2.0-beta-10"
+version = "2.0-beta-11"
 description = "Allows the dynamic launch of Jenkins agent on a Mesos cluster, depending on workload"
 sourceCompatibility = 1.8
 targetCompatibility = 1.8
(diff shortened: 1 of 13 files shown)


src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
```diff
@@ -131,12 +132,14 @@ public MesosCloud(
     this.mesosAgentSpecTemplates = mesosAgentSpecTemplates;
     this.frameworkName = frameworkName;
 
-    if (this.frameworkId == null) {
+    this.frameworkId = frameworkId;
+    if (StringUtils.isEmpty(this.frameworkId)) {
+      logger.info("Framework ID was null.");
```
vinodkone: I would log this as "Subscribing to Mesos with a new Framework ID" and actually put this after the next line so that we can log the actual ID as well. Much more helpful for debugging.

build.gradle
```diff
@@ -58,7 +58,7 @@ test {
 }
 
 group = "org.jenkins-ci.plugins"
-version = "2.0-beta-9"
+version = "2.0-beta-10"
 description = "Allows the dynamic launch Jenkins agent on a Mesos cluster, depending on workload"
```
vinodkone: s/launch/launch of/

src/main/java/org/jenkinsci/plugins/mesos/MesosApi.java
```diff
@@ -44,10 +44,50 @@
...
+ * Provides a simplified interface to Mesos through USI.
+ *
+ * <p>Each connection should be a singleton. New instance are create via {@link
+ * MesosApi#getInstance(String)}.
+ */
 public class MesosApi {
 
   private static final Logger logger = LoggerFactory.getLogger(MesosApi.class);
 
+  static HashMap<String, MesosApi> sessions = new HashMap<>();
```
vinodkone: Is this a HashMap because one could create multiple MesosCloud's from one Jenkins master?

src/main/java/org/jenkinsci/plugins/mesos/MesosCloud.java
jeschkies: I think USI is doing this already. I'm removing this line.

src/main/java/org/jenkinsci/plugins/mesos/MesosApi.java
jeschkies: Yes. The new plugin supports multiple `MesosCloud` configuration. It actually was a feature request by a customer at some point.

jeschkies:
Depends on https://github.com/mesosphere/usi/pull/133

PR #388, opened 2020-01-20
Import universe package of Jenkins.

jeschkies:


jeschkies:
Packaging will stay in https://github.com/mesosphere/dcos-jenkins-service

PR #393, opened 2020-02-27
Use Docker entrypoint.

jeschkies:
Summary:
If a Docker container is defined we pass the agent command as arguments
to the container. This way users will not have override the shell command.

Relates to https://github.com/mesosphere/dcos-jenkins-service/pull/319.

JIRA issues: DCOS_OSS-5937


jeschkies:
The failing test passes locally.

PR #394, opened 2020-04-24
Handle all QueueOfferResult types.

jeschkies:
Summary:
This will handle `enqeued`, `dropped`, `failure` and other cases. We also restart the USI scheduler flow in case the connection to Mesos is lost.

Depends on mesosphere/usi#147.

JIRA issues: D2IQ-67370


jeschkies:
We need to redefine the `MesosApi` interface so that we can test it with a mocked USI flow.

src/main/java/org/jenkinsci/plugins/mesos/api/Session.java
```diff
@@ -0,0 +1,160 @@
...
+  private static CompletableFuture<MesosClient> connectClient(
+      Protos.FrameworkInfo frameworkInfo,
+      Settings operationalSettings,
+      MesosClientSettings clientSettings,
+      Optional<CredentialsProvider> authorization,
+      ActorSystem system,
+      ActorMaterializer materializer) {
+
+    // TODO: move times and max retries into operational settings.
+    return RestartSource.onFailuresWithBackoff(
```
jeschkies, line 119: I thinks this should be move into USI. The client should always try to reconnect if the initial connection fails. Please note that this is different from restarting the scheduler flow.

PR #396, opened 2020-05-05
select default mesos slave attributes on label match partial match

mreilly82:
- In situation where developers are using Configuration as Code, and specifying mesos label in scripted DSL (ex: mesos:myDockerContainer) would require label creation for N number of docker containers to be configured in Mesos Cloud config. Was observing in this case, slave attributes were not being honored and offers were accepted regardless.
- Introduce enhancement to label matching to retrieve attributes, so in event of match on primary label configured as default (ex: mesos) for request(mesos:myDockerContainer) would match based on enhancement which first attempts to get exact match, then trims label provided to find a match also marked as default or otherwise, like previously, just return the provided label's attributes. 
- this enhancement provides to maintainers the ability to support slave info defaults in the event that users of the system want to use the default mesos attributes but would like to customize docker container being used.
- Added UTs to test previous behavior still intact, as well as testing the enhancements.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -626,9 +626,13 @@ public class JenkinsScheduler implements Scheduler {
         double requestedCpus = request.request.cpus;
         double requestedMem = (1 + JVM_MEM_OVERHEAD_FACTOR) * request.request.mem;
         // Get matching slave attribute for this label.
-        LOGGER.info("Get slave attributes for label: " + request.request.slaveInfo.getLabelString());
-        String defaultLabel = getLabelOrDefault(request);
-        JSONObject slaveAttributes = getSlaveAttributeForLabel(defaultLabel);
+        JSONObject slaveAttributes = new JSONObject();
+        String reqLabel = request.request.slaveInfo.getLabelString();
+        if(reqLabel != null && !reqLabel.isEmpty()) {
+            LOGGER.info("Get slave attributes for label: " + reqLabel);
+            String defaultLabel = getLabelOrDefault(reqLabel);
+            slaveAttributes = getSlaveAttributeForLabel(defaultLabel);
+        }
 
         if (requestedCpus <= cpus
                 && requestedMem <= mem
@@ -664,7 +668,7 @@ public class JenkinsScheduler implements Scheduler {
     private JSONObject getSlaveAttributeForLabel(String labelName) {
         LOGGER.info(String.format("Getting slave attributes for label: %s", labelName));
         List<MesosSlaveInfo> matchedSlaves = getMesosCloud().getSlaveInfos().stream()
-                .filter(slaveInfo -> slaveInfo.getLabelString() != null && slaveInfo.getLabelString().equals(labelName))
+                .filter(slaveInfo -> Objects.equals(labelName, slaveInfo.getLabelString()))
                 .collect(Collectors.toList());
 
         if (matchedSlaves.size() > 1) {
@@ -675,26 +679,25 @@ public class JenkinsScheduler implements Scheduler {
                     .stream()
                     .findFirst()
                     .map(MesosSlaveInfo::getSlaveAttributes)
-                    .orElse(null);
+                    .orElse(new JSONObject());
         }
-        return matchedSlaves.stream().findFirst().map(MesosSlaveInfo::getSlaveAttributes).orElse(null);
+        return matchedSlaves.stream().findFirst().map(MesosSlaveInfo::getSlaveAttributes).orElse(new JSONObject());
     }
 
     /**
      * Checks for label to match based upon provided label, in the event that label is not found then determine closest
      * matching label. First checks for exact match of label, on exact match return it. Secondly, will trim label (ex: mesos:myDockerContainer)
      * then checks for a match of trimmed label and returns it. If no match is found return original reqLabel.
-     * @param request
+     * @param reqLabel
      * @return label
      */
-    private String getLabelOrDefault(Request request) {
-        String reqLabel = request.request.slaveInfo.getLabelString();
+    private String getLabelOrDefault(String reqLabel) {
         boolean match = getMesosCloud().getSlaveInfos().stream().anyMatch(slaveInfo -> {
             assert slaveInfo.getLabelString() != null;
             return slaveInfo.getLabelString().equals(reqLabel);
         });
 
-        if (match || reqLabel == null) {
+        if (match) {
             return reqLabel;
         }
 
--- a/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
+++ b/src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
@@ -49,6 +49,7 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import static org.apache.mesos.Protos.ContainerInfo.Type.MESOS;
 import static org.apache.mesos.Protos.Image.Type.DOCKER;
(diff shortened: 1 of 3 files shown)


jeschkies:
Thanks for the patch @mreilly82. I will take a look tomorrow. Did you try out 2.x of the plugin? Also, I'm not sure setting the Docker image via the label is the way to go. One should use the Docker in Docker image and use Jenkins' `agent { docker { image: ... } }` declaration.

mreilly82:
Hi @jeschkies thanks! Yes, at current moment the mesos version I am using in cluster is a bit dated and wasnt able to deploy 2.x, gave an exception needing 1.9 mesos, currently using 1.5. In using the scripted pipeline DSL, where container is derived from `node("label:container"){...} `observed the labels looking to be matched based on entire label being used.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -650,6 +653,66 @@ private boolean matches(Offer offer, Request request) {
...
+     * returns attributes for SlaveInfo marked as default, otherwiese the matched label's slave attributes. Keeping with
+     * the previous behavior, will return null/empty slave attributes.
+     * @param labelName MesosSlaveInfo label string name.
+     * @return slaveAttributes as a JSONObject.
+     */
+
+    private JSONObject getSlaveAttributeForLabel(String labelName) {
+        LOGGER.info(String.format("Getting slave attributes for label: %s", labelName));
+        List<MesosSlaveInfo> matchedSlaves = getMesosCloud().getSlaveInfos().stream()
+                .filter(slaveInfo -> slaveInfo.getLabelString() != null && slaveInfo.getLabelString().equals(labelName))
```
jeschkies, line 667:
```suggestion
                .filter(slaveInfo -> labelName.equals(slaveInfo.getLabelString()))
```
should avoid the null pointer exception.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -650,6 +653,66 @@ private boolean matches(Offer offer, Request request) {
...
+                .collect(Collectors.toList());
+
+        if (matchedSlaves.size() > 1) {
+            LOGGER.info(String.format("Multiple matches for label %s found, matches: %o", labelName, matchedSlaves.size()));
+            LOGGER.info("For multiple matches, checking for Slave Info marked as default");
+            return matchedSlaves.stream()
+                    .filter(MesosSlaveInfo::isDefaultSlave).collect(Collectors.toList())
+                    .stream()
```
jeschkies, line 675: Is this required?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -650,6 +653,66 @@ private boolean matches(Offer offer, Request request) {
...
+                    .findFirst()
+                    .map(MesosSlaveInfo::getSlaveAttributes)
+                    .orElse(null);
```
jeschkies, line 678: Not sure if we want to return `null`.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -650,6 +653,66 @@ private boolean matches(Offer offer, Request request) {
...
+     * matching label. First checks for exact match of label, on exact match return it. Secondly, will trim label (ex: mesos:myDockerContainer)
+     * then checks for a match of trimmed label and returns it. If no match is found return original reqLabel.
+     * @param request
+     * @return label
+     */
+    private String getLabelOrDefault(Request request) {
+        String reqLabel = request.request.slaveInfo.getLabelString();
+        boolean match = getMesosCloud().getSlaveInfos().stream().anyMatch(slaveInfo -> {
+            assert slaveInfo.getLabelString() != null;
+            return slaveInfo.getLabelString().equals(reqLabel);
```
jeschkies, line 694:
```suggestion
            reqLabel.equals(slaveInfo.getLabelString());
```

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
```diff
@@ -650,6 +653,66 @@ private boolean matches(Offer offer, Request request) {
...
+        });
+
+        if (match || reqLabel == null) {
+            return reqLabel;
```
jeschkies, line 698: I don't quite understand. Why are we returning null here when `reqLabel` is `null`?

src/test/java/org/jenkinsci/plugins/mesos/JenkinsSchedulerTest.java
```diff
@@ -204,6 +205,72 @@ public void testDeclineOffersWithRequestsInQueue() throws Exception {
         Mockito.verify(driver, Mockito.never()).suppressOffers();
     }
 
+    @Test
+    public void testSlaveAttributesFromLabel() throws Exception {
```
jeschkies, line 209: Nice! I just wanted to ask for tests 🙂

src/test/java/org/jenkinsci/plugins/mesos/JenkinsSchedulerTest.java
```diff
@@ -204,6 +205,72 @@ public void testDeclineOffersWithRequestsInQueue() throws Exception {
...
+        List<MesosSlaveInfo> mesosSlaveInfos = mockSlaveInfos();
+        Mockito.when(mesosCloud.getSlaveInfos()).thenReturn(mesosSlaveInfos);
+        MesosComputer computer = Mockito.mock(MesosComputer.class);
+        Mockito.when(jenkins.getComputers()).thenReturn(new Computer[] { computer });
+        SchedulerDriver driver = Mockito.mock(SchedulerDriver.class);
+        jenkinsScheduler.setDriver(driver);
+
+        //check it
+        jenkinsScheduler.resourceOffers(driver, offers);
+        Mockito.verify(driver, Mockito.atLeastOnce()).declineOffer(rejectedOffer.getId(), Protos.Filters.newBuilder().setRefuseSeconds(5).build());
```
jeschkies, line 227: Which offer are we expecting to be accepted?

src/test/java/org/jenkinsci/plugins/mesos/JenkinsSchedulerTest.java
```diff
@@ -573,6 +748,30 @@ public void getFrameworkId() {
                 .build();
     }
 
+    private Protos.Offer createOfferWithAttributes(String name, String value) {
```
jeschkies, line 751: Could you add documentation?

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
mreilly82: I believe so, in testing I found that one could define multiple labels in config with same name. In this case, I check for one of the matches to be marked as default, otherwise in event of single match label proceeds with retrieving slave attributes of the matched label and if not present an empty JSONObject returned.

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
mreilly82: Right, proceeded to return JSONObject instead of null

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
mreilly82: 👍

src/main/java/org/jenkinsci/plugins/mesos/JenkinsScheduler.java
mreilly82: 👍 handling reqLabel being null will be handled prior to this method being invoked.

jeschkies:
Thanks for the changes. It seems `JenkinsSchedulerTest` is failing.

mreilly82:
Hi @jeschkies I checked Test Result and it seems not all tests ran, re-triggered by push(no changes) and all tests ran this time. Thanks 👍

jeschkies:
Thanks!

PR #400, opened 2020-05-15
Fix DockerAgentTest.

jeschkies:
Summary:
The previous image would try to start a Docker daemon in a Docker container which itself is
running in a Docker container on the CI. This could work but just takes too much configuration.
We are running a simple client instead.


jeschkies:
It fails now with `Error starting daemon: Devices cgroup isn't mounted`. I did test it on a Linux machine. Maybe we should just use a simple Docker image without trying Docker in Docker.

PR #406, opened 2020-06-12
Use no delay provision startegy.

jeschkies:
Summary:
The current provision startegy considers the pending tasks in Mesos.
This number can be out dated. We should only consider the state from
Jenkins and launch the Jenkins agent immediately without a delay.

JIRA issues: D2IQ-68723


jeschkies:
We probably should also bump USI.

kaiwalyajoshi:
Ok, we're following [NoDelayProvisionerStrategy](https://github.com/jenkinsci/kubernetes-plugin/blob/master/src/main/java/io/jenkins/plugins/kubernetes/NoDelayProvisionerStrategy.java) from kubernetes-plugin with minor changes to logging.

PR #421, opened 2024-02-02
Update to Gradle 8.5.0

jeschkies:


jeschkies:
I've tried updating Gradle https://github.com/jenkinsci/mesos-plugin/pull/421 but haven't found the time since. I don't think this repo is much alive anymore :/

